PR #8, opened 2021-09-30
Handle chunking, append/set outputs

zhan-airplane:


package.json
```diff
@@ -24,5 +24,8 @@
     "np": "^7.5.0",
     "ts-jest": "^26.5.6",
     "typescript": "^4.2.4"
+  },
+  "dependencies": {
+    "@types/uuid": "^8.3.0"
```
colinking, line 29: `@types` dependencies should be in `devDependencies` since they're only needed during development (since the TS is transpiled to JS before releasing to npm). You'll also need to add `uuid` to `dependencies` which'll include the actual code for the `uuid` package (`@types/uuid` is just the types [from DefinitelyTyped](https://github.com/DefinitelyTyped/DefinitelyTyped/tree/master/types/uuid))

index.ts
```diff
@@ -1,32 +1,91 @@
...
+ * @example
+ *   setOutput("Task completed successfully.")
+ *
+ * @example
+ *   setOutput({"values": {"error": "This is an error message"}})
+ *
+ * @example
+ *   setOutput("A different error message", "values.error")
+ */
+function setOutput(value: any, path?: string) {
```
colinking, line 48: Same comment as: https://github.com/airplanedev/go-sdk/pull/6#discussion_r719642736

Though we'll need a separate TS-based implementation.

index.ts
```diff
@@ -1,32 +1,91 @@
...
+ * To learn more about Airplane outputs, see the docs: https://docs.airplane.dev/reference/outputs
+ *
+ * @example
+ *   appendOutput("Task completed successfully.")
+ *
+ * @example
+ *   appendOutput({"name": "Carolyn", "occupation": "agent"}, "rows")
+ *
+ */
+function appendOutput(value: any, path?: string) {
```
colinking, line 68: Could we mirror the `(pathOrValue, path?)` layout of `output`? It reads more naturally if the path comes first, when providing both a path and value.

index.test.ts
```diff
@@ -53,6 +53,142 @@ test("output values with keys", () => {
...
+    `airplane_output_append:abc["def"].ghi[0] true`,
+    `airplane_output_append:abc["def"].ghi[0] false`,
+    `airplane_output_append:abc["def"].ghi[0] 123`,
+    `airplane_output_append:abc["def"].ghi[0] 123.456`,
+    `airplane_output_append:abc["def"].ghi[0] ["hello","world"]`,
+    `airplane_output_append:abc["def"].ghi[0] {"catchphrase":"that's too much, man!"}`,
+  ]);
+});
+
+test("chunking", () => {
```
colinking, line 146: 🔥

colinking (changes requested):

index.ts
zhan-airplane: Hmm I'm not a huge fan actually, though I'm willing to be convinced - I think it makes the arg names + types a little less informative.

Also (might be wrong here), if someone happens to inadvertently pass `undefined` through to the second argument, it would overwrite the JSON output with the path string right? I'm not sure I like that risky behavior either haha

colinking: > Also (might be wrong here), if someone happens to inadvertently pass undefined through to the second argument, it would overwrite the JSON output with the path string right? I'm not sure I like that risky behavior either haha

Good point, so if we tried to handle this by looking if path is `undefined | null`, then we couldn't figure this case out:

```ts
setOutput("name", undefined)
```

Did they want to log `name=undefined`, or `<root>="name"`? I'd lean towards just saying that if you pass two args (by checking `arguments.length` like before) then we assume you want `name=undefined`.

But yeah, agreed this could become a footgun. Keeping (value, path) sgtm then!

index.ts
```diff
@@ -1,32 +1,113 @@
...
+ * To learn more about Airplane outputs, see the docs: https://docs.airplane.dev/reference/outputs
+ *
+ * @example
+ *   setOutput("Task completed successfully.")
+ *
+ * @example
+ *   setOutput({"values": {"error": "This is an error message"}})
+ *
+ * @example
+ *   setOutput("A different error message", "values", "error")
```
colinking, line 46:
```suggestion
 *   setOutput("Received a 500 from the upstream API", "error")
```

index.ts
```diff
@@ -1,32 +1,113 @@
...
+ * This sets the Airplane output (`value`). If a path is provided, it sets the
+ * portion of the output described by the path. For path syntax, please see the docs.
+ *
+ * To learn more about Airplane outputs, see the docs: https://docs.airplane.dev/reference/outputs
+ *
+ * @example
+ *   setOutput("Task completed successfully.")
+ *
+ * @example
+ *   setOutput({"values": {"error": "This is an error message"}})
```
colinking, line 43: Let's use real world examples. It makes it easier to connect with how you would use in practice :)

```suggestion
 *   setOutput({"error": "Received a 500 from the upstream API"})
```

PR #18, opened 2022-04-18
Add `airplane.execute` method

colinking:
This PR adds a new `airplane.execute` method that can be used to execute other tasks by slug. For example:

```ts
import airplane from './index'

export default async function() {
  const run = await airplane.execute<string>("capitalize", {
    input: "hello world"
  })

  return run
}
```


colinking:
* main
  * **PR #18** <a href="https://app.graphite.dev/github/pr/airplanedev/node-sdk/18" target="_blank">Graphite</a>  (this PR)

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/airplanedev/node-sdk/18?utm_source=stack-comment).

tasks.ts
```diff
@@ -0,0 +1,61 @@
...
+  paramValues: Input
+  status: RunStatus
+  output: Output
+}
+
+export async function execute<Output = unknown>(slug: string, params?: Record<string, any> | undefined | null): Promise<Run<typeof params, Output>> {
+  const fetcher = new Fetcher({
+    host: process.env.AIRPLANE_API_HOST ?? "",
+    // https://github.com/airplanedev/airport/pull/2027
+    token: process.env.AIRPLANE_TOKEN ?? process.env.AIRPLANE_RUN_AUTHN_TOKEN ?? "",
```
francescaleung, line 27: should these just be a part of Fetcher?

tasks.ts
```diff
@@ -0,0 +1,61 @@
...
+  })
+
+  const { runID } = await fetcher.post<{
+    runID: string
+  }>("/v0/tasks/execute", {
+    slug,
+    paramValues: params ?? {},
+  })
+
+  // Poll until the run is ready:
```
francescaleung, line 37: ready -> done

poll.ts
```diff
@@ -0,0 +1,36 @@
...
+  constructor(opts: PollOptions) {
+    this.opts = opts
+  }
+
+  async run<Output = unknown>(fn: PollFunction<Output>): Promise<Output> {
+    return new Promise(async (resolve, reject) => {
+      const fnw = async () => {
+        try {
+          const out = await fn()
+          if (out != null) {
```
francescaleung, line 23: `!==`?

poll.ts
```diff
@@ -0,0 +1,36 @@
...
+            return resolve(out)
+          }
+        } catch (err) {
+          return reject(err)
+        }
+  
+        setTimeout(fnw, this.opts.delayMs)
```
francescaleung, line 30: is there a need to worry about cleanup here? Is there a reason we'd want to cancel the poller out-of-band wrt poller delay? I kind of don't think so but figured I'd bring it up just in case.

francescaleung (approved):

tasks.ts
colinking: No, this allows us to provide different values when testing `Fetcher`.

poll.ts
colinking: Not that I know of. Process shutdown should work fine, which is the main out-of-band signal that comes to mind.

PR #22, opened 2022-04-19
Configure Jest to handle ESM deps

colinking:
Updates our Jest config to handle ESM dependencies like `node-fetch`.

PR #21, opened 2022-04-19
Run tests on all node versions

colinking:
Updates our test suite to run on all versions of Node.js that are currently supported on Airplane, [except Node 12 and 15](https://linear.app/airplane/issue/AIR-3536/deprecate-node-12-and-15). It also adds [Node 18](https://linear.app/airplane/issue/AIR-3549/support-node-18) given we'll be adding support for it to Airplane soon.

We don't include Node 12 support here since `node-fetch` uses the `node:` protocol (e.g. `node:http`) which is only available in ESM mode in Node 12. For folks using the Node 12 runtime, they can continue to use `v0.1.2`.

Based on: https://github.com/sindresorhus/got/blob/main/.github/workflows/main.yml

PR #23, opened 2022-04-19
Add test suite for `airplane.execute`

colinking:
Adds test cases that cover the `airplane.execute` method, along with the `Poller` and `Fetcher` classes.


colinking:
* main
  * **PR #21** <a href="https://app.graphite.dev/github/pr/airplanedev/node-sdk/21" target="_blank">Graphite</a> 
    * **PR #22** <a href="https://app.graphite.dev/github/pr/airplanedev/node-sdk/22" target="_blank">Graphite</a> 
      * **PR #23** <a href="https://app.graphite.dev/github/pr/airplanedev/node-sdk/23" target="_blank">Graphite</a>  (this PR)

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/airplanedev/node-sdk/23?utm_source=stack-comment).

fetch.ts
```diff
@@ -7,22 +7,31 @@ import { version } from "./package.json";
...
+    this.host = opts.host;
+
     if (!opts.token) {
       throw new Error("expected an authentication token");
     }
+    this.token = opts.token;
 
-    this.opts = opts;
+    const defaultRetryDelay: FetchOptions["retryDelay"] = (attempt) => {
+      return [0, 100, 200, 400, 600, 800, 1000][attempt - 1] ?? 1000;
```
francescaleung, line 32: attempt starts at 0, does that mean the first retry happens after 1s?

tasks.test.ts
```diff
@@ -0,0 +1,51 @@
+import nock from "nock";
+
+import { execute, RunStatus } from "./tasks";
+
+const OLD_ENV = process.env;
+afterAll(() => {
+  process.env = OLD_ENV;
```
francescaleung, line 7: if we're just rewriting env for tests, then the Fetcher can default to pulling host/token from the env, right? we can add an explicit override option if necessary?

francescaleung (approved):

fetch.ts
colinking: Good catch! Fixed

tasks.test.ts
colinking: We're rewriting them just for the `airplane.execute` tests, not the `Fetcher` tests. Otherwise we'd need to mock the env in both places. Down the road, will likely move the `process.env` logic out of the `execute` method which would allow us to remove this logic from here.

PR #24, opened 2022-04-21
Fix builds

colinking:


colinking:
* main
  * **PR #24** <a href="https://app.graphite.dev/github/pr/airplanedev/node-sdk/24" target="_blank">Graphite</a>  (this PR)

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/airplanedev/node-sdk/24?utm_source=stack-comment).

francescaleung (approved):
possibly dumb question: why do we need a different tsconfig file?

colinking:
> possibly dumb question: why do we need a different tsconfig file?

It's a good question! I tried to get it working that way. `include/exclude` in `tsconfig.json` determine which files to compile, but you can also set them from the CLI e.g. `tsc index.ts`. The issue is that you can only do one or the other -- if you pass files to the `tsc` CLI then it ignores your `tsconfig.json`. You also can't pass both, e.g. `tsc -p tsconfig.json index.ts`.

PR #28, opened 2022-05-17
Use cross fetch instead of node-fetch

leeweisberger:
Use cross fetch which works in browser (uses [GitHub's WHATWG Fetch polyfill](https://github.com/github/fetch)) or node (uses node-fetch).
This should just wrap the two libraries and not do anything extra. Later, we might want to consider writing our own impl that just uses browser fetch instead of the polyfill to reduce code size

### DIFF
--- a/package.json
+++ b/package.json
@@ -23,10 +23,9 @@
   "devDependencies": {
     "@babel/core": "7.17.9",
     "@babel/plugin-transform-modules-commonjs": "7.17.9",
-    "@types/isomorphic-fetch": "^0.0.36",
     "@types/jest": "27.4.1",
     "@types/node": "16.11.7",
-    "@types/node-fetch": "^2.6.1",
+    "@types/node-fetch": "2.6.1",
     "@types/uuid": "8.3.4",
     "@typescript-eslint/eslint-plugin": "5.18.0",
     "@typescript-eslint/parser": "5.18.0",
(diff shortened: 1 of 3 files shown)


package.json
```diff
@@ -52,7 +54,7 @@
   },
   "dependencies": {
     "fetch-retry": "5.0.2",
-    "node-fetch": "3.2.3",
+    "isomorphic-fetch": "^3.0.0",
```
colinking, line 57: pin this for renovate

colinking (changes requested):
Did a quick search. Looks like `cross-fetch` is a better option: https://github.com/lquixada/cross-fetch

`isomorphic-unfetch` is using fairly old versions of `node-fetch`

leeweisberger:
> Did a quick search. Looks like `cross-fetch` is a better option: https://github.com/lquixada/cross-fetch
> 
> `isomorphic-unfetch` is using fairly old versions of `node-fetch`

Fine with this!

colinking (approved):
LGTM. Any thoughts on how we can include unit tests for the browser env?

colinking:
Also @leeweisberger make sure to test deploying a Node task with this version of the SDK!

leeweisberger:
> LGTM. Any thoughts on how we can include unit tests for the browser env?

In web, we use https://mswjs.io/ which mocks at the network layer and then polyfill in `whatwg-fetch` during tests so that we actually make fetch requests at the server. We could set that up

PR #29, opened 2022-06-03
Support passing in authn to execute as params

leeweisberger:
Support using `execute` in non-node environments without depending on environment variables. I'm [currently working around this](https://github.com/airplanedev/views/blob/1c80a979629a5cd432cef2ef76a50d34bbab0433/examples/executeTask/src/main.tsx#L10) in views, but it's icky.

Other than being able to pass authn/host directly to `execute`, I created a client that we can pass authn information to once and then continue to call execute - this seems like a more standard pattern, but definitely not necessary.

We also want to support apiKey authn because the CLI supports it.

FIXES AIR-3913

### DIFF
--- a/fetch.ts
+++ b/fetch.ts
@@ -7,14 +7,16 @@ import { version } from "./package.json";
 
 export type FetchOptions = {
   host: string;
-  token: string;
+  token?: string;
+  apiKey?: string;
   retryDelay?: (attempt: number) => number;
 };
 
 // Fetcher is a wrapper around node-fetch with reasonable defaults that understands the Airplane API format.
 export class Fetcher {
   private host: string;
-  private token: string;
+  private token?: string;
+  private apiKey?: string;
   private fetch: ReturnType<typeof withFetchRetries>;
   private retryDelay: FetchOptions["retryDelay"];
 
@@ -23,11 +25,12 @@ export class Fetcher {
       throw new Error("expected an api host");
     }
     this.host = opts.host;
-
-    if (!opts.token) {
-      throw new Error("expected an authentication token");
-    }
     this.token = opts.token;
+    this.apiKey = opts.apiKey;
+
+    if (!this.token && !this.apiKey) {
+      throw new Error("expected an authentication method");
+    }
 
     const defaultRetryDelay: FetchOptions["retryDelay"] = (attempt) => {
       return [0, 100, 200, 400, 600, 800, 1000][attempt] ?? 1000;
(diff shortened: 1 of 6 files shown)


fetch.ts
```diff
@@ -75,14 +78,20 @@ export class Fetcher {
...
+    const headers: HeadersInit = {
+      "X-Airplane-Client-Kind": "sdk/node",
+      "X-Airplane-Client-Version": version,
+    };
+    if (this.token) {
+      headers["X-Airplane-Token"] = this.token;
+    }
+    if (this.apiKey) {
+      headers["X-Airplane-API-Key"] = this.apiKey;
+    }
```
colinking, line 90: We shouldn't set both in the API response (e.g. which should the API use?). Let's error if both are set.

package.json
```diff
@@ -7,7 +7,8 @@
   "author": "hello@airplane.dev",
   "license": "MIT",
   "scripts": {
-    "test": "rm -rf dist && jest --testTimeout 500",
+    "test": "jest --testTimeout 500",
+    "test:watch": "jest --watch",
```
colinking, line 11: Add `--testTimeout`?

tasks.ts
```diff
@@ -20,13 +20,24 @@ export type Run<Input = unknown, Output = unknown> = {
...
+
 export const execute = async <Output = unknown>(
   slug: string,
-  params?: Record<string, unknown> | undefined | null
+  params?: Record<string, unknown> | undefined | null,
+  opts?: ExecuteOptions
 ): Promise<Run<typeof params, Output>> => {
+  const host = process?.env?.AIRPLANE_API_HOST || opts?.host || "";
+  const token = process?.env?.AIRPLANE_TOKEN || opts?.token;
+  const apiKey = process?.env?.AIRPLANE_API_KEY || opts?.apiKey;
```
colinking, line 36: Let's flip this so `opts` can override `env`

tasks.ts
```diff
@@ -61,3 +72,19 @@ export const execute = async <Output = unknown>(
     };
   });
 };
+
+export type TaskClientOptions = {
+  host?: string;
+  token?: string;
+  apiKey?: string;
+};
```
colinking, line 80: nit: add a newline here

tasks.ts
```diff
@@ -61,3 +72,19 @@ export const execute = async <Output = unknown>(
...
+export class TaskClient {
```
colinking, line 81: Do we need this? You can already get the same behavior through execute options. I'd prefer we have one way to do operations from the SDK.

colinking (approved):
LGTM assuming we do not export the TaskClient. Re-request if you want to keep it.

tasks.ts
leeweisberger: I can remove. The only reason I added this is so views doesn't have to pass in the authn functions on every call to execute, instead creating a client once that handles authn and then calling execute. But was a very small optimization

colinking: Makes sense. Would it make sense to wrap the `execute` function in Views instead?

PR #31, opened 2022-06-06
Propagate env to child runs

ameyakhare:
We want to pass down the run environment to child runs so they execute in the correct environment.

### DIFF
--- a/fetch.ts
+++ b/fetch.ts
@@ -8,6 +8,7 @@ import { version } from "./package.json";
 export type FetchOptions = {
   host: string;
   token: string;
+  envID: string;
   retryDelay?: (attempt: number) => number;
 };
 
@@ -15,6 +16,7 @@ export type FetchOptions = {
 export class Fetcher {
   private host: string;
   private token: string;
+  private envID: string;
   private fetch: ReturnType<typeof withFetchRetries>;
   private retryDelay: FetchOptions["retryDelay"];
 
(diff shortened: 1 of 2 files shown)


fetch.ts
```diff
@@ -29,6 +31,11 @@ export class Fetcher {
     }
     this.token = opts.token;
 
+    if (!opts.envID) {
+      throw new Error("expected an environmentID");
```
colinking, line 35:
```suggestion
      throw new Error("expected an environment ID");
```

tasks.ts
```diff
@@ -27,6 +27,7 @@ export const execute = async <Output = unknown>(
   const fetcher = new Fetcher({
     host: process.env.AIRPLANE_API_HOST ?? "",
     token: process.env.AIRPLANE_TOKEN ?? "",
+    envID: process.env.AIRPLANE_ENV_ID ?? "",
```
colinking, line 30: @leeweisberger made some related changes in https://github.com/airplanedev/node-sdk/pull/29 -- make sure to pull in `main`

colinking (approved):
LGTM, assuming there will be a separate validation PR for API.

tasks.ts
```diff
@@ -24,6 +24,7 @@ export type ExecuteOptions = {
   host?: string;
   token?: string;
   apiKey?: string;
+  envID?: string;
```
leeweisberger, line 27: Thanks! I was going to do this but forgot.

ameyakhare: :widepeepohappy:

PR #32, opened 2022-06-07
Process does not exist in non-node

leeweisberger:
This code returns a reference error in non node environemnts because process does not exist globally. We need to use a `typeof` check which doesn't throw a reference error when using process.


ameyakhare (approved):
🚀

PR #34, opened 2022-06-08
Add support for workflow task execution

calufornia:
This PR adds the first workflow-specific functionality to the Node SDK by extending `airplane.execute` to workflow tasks. It mimics the behavior of the current `airplane.execute` functionality for standard functions in that it kicks off execution of a "child" task. However, the mechanism that it uses is different - instead of polling, it:
1. Executes a Temporal activity that hits the Airplane API to start execution of a task with the given slug
2. Registers a Temporal signal that a Temporal client (i.e. our API) can trigger
3. Waits until a signal has been received with the appropriate run status before continuing the workflow.

Note that this relies on the Temporal worker that is included at _build_ time, which means all of the user's code is wrapped in a Temporal workflow. This means that we don't currently support env vars in workflow tasks that use the Node SDK, as [env vars result in nondeterminism in workflows](https://deploy-preview-1205--mystifying-fermi-1bc096.netlify.app/application-development-guide/#environment-variables). We can add a Temporal activity (which does have access to env vars) that populates a map of env vars - there are also plans on extracting the Temporal worker code from build time to run time (Phase 1).

cc @yolken-airplane 

Testing done by creating a task that referenced the Node SDK locally:
```
import airplane from "./index";

// Put the main logic of the task in this function.
export default async function(params) {
  return await airplane.execute("rest_task", undefined, {
    apiKey: <omitted>,
    envID: "env20220125z7pz51c",
    runtime: "workflow",
  });
}
```

Then, manually signaling the workflow using tctl:
```
tctl --ns 1jq9NDeTcTmyqqXAjEsk7V0wXaV wf s -w run20220608zjsr4509vgv -n run20220608zcbhj0erdwe -i '{"taskID": "trv20220606z3pona1obnn", "paramValues": null, "status": "Succeeded"}'
```


tasks.ts
```diff
@@ -25,54 +35,78 @@ export type ExecuteOptions = {
...
+): Promise<Run<typeof params, Output>> => {
+  const runID = await executeTaskActivity(slug, params, opts);
+
+  let taskID = "";
+  let status: RunStatus = RunStatus.NotStarted;
+  let paramValues = undefined;
+  // Register termination signal for the workflow. We ensure signal name uniqueness by utilizing the run ID of the task
+  // being executed, since a workflow task may execute an arbitrary number of other tasks.
+  const taskSignal = wf.defineSignal(runID);
+  // @ts-ignore
```
calufornia, line 93: I'm not sure why TypeScript complains about the handler signature - it works locally (payload is not null), and they also have examples that reference a payload: https://docs.temporal.io/typescript/workflows#define-signals-and-queries-dynamically

activities.ts
```diff
@@ -0,0 +1,23 @@
...
+import { executeTask, fetchTaskOutput, getFetcher } from "./tasks_utils";
+
+// Recommended activity factory by Temporal: https://docs.temporal.io/typescript/activities/#important-design-patterns
+export const registerActivities = () => ({
+  executeTaskActivity: async (
+    slug: string,
+    params: Record<string, unknown> | undefined | null,
+    opts?: ExecuteOptions
+  ): Promise<string> => {
+    // We re-initialize the fetcher in each activity because there is an issue with Temporal proxy activities where it
```
calufornia, line 11: For more context, when I do something like
`console.log(fetcher.get)` inside the activity, it outputs an async function (expected). However, if I were to have an activity such as:

```
initializeFetcherActivity: async(
  opts?: ExecuteOptions
): Promise<Fetcher> => {
  return getFetcher(opts);
}
```

and then call the activity in the workflow (via `proxyActivities`) with something like:

```
const fetcher = await initializeFetcherActivity(opts);
console.log(fetcher.get)
```
The second line outputs `undefined`, as if the class method has been stripped from it. There might something obvious I'm missing here.

tasks.ts
```diff
@@ -25,54 +35,78 @@ export type ExecuteOptions = {
...
+  wf.setHandler(taskSignal, (payload) => {
+    taskID = payload.taskID;
+    paramValues = payload.paramValues;
+    status = payload.status;
   });
+
+  // Defer workflow execution until the task has been completed.
+  await wf.condition(() => terminalStatuses.includes(status));
+
+  const output = await fetchTaskOutputActivity<Output>(runID, opts);
```
colinking:
```suggestion
  const output = await getRunOutputActivity<Output>(runID, opts);
```

tasks.ts
```diff
@@ -25,54 +35,78 @@ export type ExecuteOptions = {
...
+  params: Record<string, unknown> | undefined | null,
+  opts?: ExecuteOptions
+): Promise<Run<typeof params, Output>> => {
+  const runID = await executeTaskActivity(slug, params, opts);
+
+  let taskID = "";
+  let status: RunStatus = RunStatus.NotStarted;
+  let paramValues = undefined;
+  // Register termination signal for the workflow. We ensure signal name uniqueness by utilizing the run ID of the task
+  // being executed, since a workflow task may execute an arbitrary number of other tasks.
```
colinking, line 90: Wouldn't we need a prefix here to guarantee uniqueness? In case we introduce other signals that are runID-based.

tasks.ts
```diff
@@ -25,54 +35,78 @@ export type ExecuteOptions = {
...
+  const taskSignal = wf.defineSignal(runID);
+  // @ts-ignore
+  wf.setHandler(taskSignal, (payload) => {
```
colinking, line 93: Add a TS type for `payload` -- this'll help define what this signal API is, too.

tasks.ts
colinking: What error do you get?

tasks.ts
```diff
@@ -1,5 +1,15 @@
...
+import { proxyActivities } from "@temporalio/workflow";
+
+import type { registerActivities } from "./activities";
 import { Poller } from "./poll";
+import { executeTask, fetchTaskOutput, getFetcher } from "./tasks_utils";
+
+const { executeTaskActivity, fetchTaskOutputActivity } = proxyActivities<
+  ReturnType<typeof registerActivities>
+>({
+  scheduleToCloseTimeout: "10m",
```
colinking, line 11: Can probably reduce this to something just over 60s (our API timeout) -- maybe `75s` or `120s`? Also let's set this on `startToClose` instead: https://docs.temporal.io/blog/activity-timeouts/#written-version-tldr

colinking: If you haven't already, also double-check if we need to adjust any default timeouts or other activity config

colinking (approved):
AD during standup:
1. Let's solidify a plan on how these activities are registered work the worker. If we just define them in the worker, do we need a separate PR to update our builder to look for `activities.js` from the SDK? If we depend on a specific file name and location, document that in `activities.ts` here so we don't mistakenly move/rename it.
2. Let's find a way to pass through the runtime/envID/apiKey automatically, possibly by mocking `process.env` in the workflow isolate. I'm +1 on adding `opts.runtime` temporarily, as long as we plan to remove it soon after.

activities.ts
colinking: Hmm, Temporal might not support serializing classes? Wonder if we need to declare specific methods to do that. Recreating the fetcher seems reasonable.

tasks.ts
colinking: Just tested locally, you need to add a type to `defineSignal`:

CleanShot 2022-06-08 at 16 42 24

index.ts
```diff
@@ -1,6 +1,7 @@
 import { appendOutput, setOutput } from "./output";
 import { execute } from "./tasks";
 export { ExecuteOptions } from "./tasks";
+export { registerActivities } from "./activities";
```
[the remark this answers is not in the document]
calufornia: Yeah agree that this is not the most ideal - we need this so that that the worker shim can register the activities: https://github.com/airplanedev/lib/blob/main/pkg/build/workflow/worker-and-activity-shim.js#L72, so that it can execute them. This is something else that gets fixed by moving worker creation to the runtime.

tasks.ts
calufornia: `error TS2345: Argument of type '(payload: any) => void' is not assignable to parameter of type '() => void | Promise<void>'.` - both in the IDE and also caught by `tasks.test.ts`.

PR #35, opened 2022-06-08
Add builtins.sql.query function

ameyakhare:
Adds a function that calls the execute api after structuring inputs correctly.

OQ:
What do we want resource structs to look like?
What should import of this function look like?

### DIFF
--- a/builtins/sql.ts
+++ b/builtins/sql.ts
@@ -1,0 +1,17 @@
+import { Run, execute } from "tasks";
+
+export enum TransactionMode {
+  Auto = "auto",
+  ReadOnly = "readOnly",
+  ReadWrite = "readWrite",
+  None = "none",
+}
+
+export const query = async <Output = unknown>(
+  query: string,
+  db: string,
+  queryArgs?: Record<string, unknown> | undefined | null,
+  transactionMode: TransactionMode = TransactionMode.Auto
+): Promise<Run<Record<string, unknown> | undefined | null, Output>> => {
+  return execute("airplane:sql_query", { query, queryArgs, transactionMode }, { db });
+};
--- a/tasks.ts
+++ b/tasks.ts
@@ -30,6 +30,7 @@ export type ExecuteOptions = {
 export const execute = async <Output = unknown>(
   slug: string,
   params?: Record<string, unknown> | undefined | null,
+  resources?: Record<string, string> | undefined | null,
   opts?: ExecuteOptions
 ): Promise<Run<typeof params, Output>> => {
   const env = typeof process === "undefined" ? {} : process?.env;
@@ -49,6 +50,7 @@ export const execute = async <Output = unknown>(
   }>("/v0/tasks/execute", {
     slug,
     paramValues: params ?? {},
+    resources: resources ?? {},
   });
 
   // Poll until the run terminates:
--- a/tsconfig.json
+++ b/tsconfig.json
@@ -9,7 +9,8 @@
     "forceConsistentCasingInFileNames": true,
     "outDir": "dist",
     "declaration": true,
-    "resolveJsonModule": true
+    "resolveJsonModule": true,
+    "baseUrl": "."
   },
   "exclude": ["node_modules", "dist"]
 }


builtins/sql.ts
```diff
@@ -0,0 +1,21 @@
+import { Run, execute } from "tasks";
+
+export enum TransactionMode {
+  Auto = "auto",
+  ReadOnly = "readOnly",
+  ReadWrite = "readWrite",
+  None = "none",
+}
+
+export const sql = {
```
colinking, line 10: Export as `default` so you can import as `import sql from 'airplane/builtins/sql`

```suggestion
export default {
```

builtins/sql.ts
```diff
@@ -0,0 +1,21 @@
...
+  TransactionMode,
+
+  query: async <Output = unknown>(
+    query: string,
+    db: string,
```
colinking, line 15: This doesn't work with aliases; you will need to look up the alias to get either an ID or slug.

colinking, line 15: Separately, I dislike APIs with sequential parameters of the same type. It's hard to tell you've mixed up the order and there's usually no "obvious" order.

```ts
airplane.sql.query("SELECT 1", "db")
airplane.sql.query("db", "SELECT 1")
```

If one of these is a `Resource` type, that's easier:

```ts
airplane.sql.query(
	airplane.resources.get("db"),
	"SELECT 1"
)
```

Another alternative would be mounting (or just aliasing?) these functions under a resource object:

```ts
airplane.resources.postgresql("db").query("SELECT 1")
```

Thoughts?

builtins/sql.ts
```diff
@@ -0,0 +1,21 @@
...
+    queryArgs?: Record<string, unknown> | undefined | null,
+    transactionMode: TransactionMode = TransactionMode.Auto
+  ): Promise<Run<Record<string, unknown> | undefined | null, Output>> => {
```
colinking, line 18: Run parameter values are a bit ... weird... for stdAPI. If you look in the runs table, `param_values = {}` and the actual params are in `std_api_request.request`.

For stdAPI, let's copy that value over into the `paramValues` we return so it is consistent with other task runs.

tasks.ts
```diff
@@ -43,14 +43,15 @@ export type ExecuteOptions = {
 export const execute = async <Output = unknown>(
   slug: string,
   params?: Record<string, unknown> | undefined | null,
+  resources?: Record<string, string> | undefined | null,
```
colinking, line 46: Don't expose `resources` from `airplane.execute` -- it's not supported for non builtin executions.

index.ts
```diff
@@ -1,10 +1,12 @@
-import { appendOutput, setOutput } from "./output";
-import { execute } from "./tasks";
-export { ExecuteOptions } from "./tasks";
-export { registerActivities } from "./activities";
+export { registerActivities } from "activities";
+import { sql } from "builtins/sql";
```
colinking, line 2: Let's drop the `builtins` folder so that you can import as

```
import sql from 'airplane/sql`

sql.query(...)
```

colinking (changes requested):
This is looking good!

builtins/sql.ts
ameyakhare: the alias for all builtins tasks is known beforehand (ie `db` for sql), so I'm unsure why this wouldn't work. also, we'll have to translate the identifier if its a slug to the id, but we should probably do that on the backend and accept both here

ameyakhare: chose option 1. since we accept a list of resources, if we go option 2 I feel like that interface might start looking different to when we start passing in multiple resources to a stdapi function

colinking (approved):

builtins/sql.ts
colinking: Ah, so the execution API will support aliases, slugs, and IDs? What happens when there's a conflict? e.g. an alias for resource 2 is the slug of resource 1?

With `AIRPLANE_RESOURCES`, you could look up an alias -> ID or slug easily enough. I'd opt to keep it simple for now.

PR #38, opened 2022-06-10
Add codeql and dependency-review

joshma:


.github/workflows/codeql.yml
```diff
@@ -0,0 +1,65 @@
...
+    runs-on: ubuntu-latest
+    permissions:
+      actions: read
+      contents: read
+      security-events: write
+
+    strategy:
+      fail-fast: false
+      matrix:
+        language: ["javascript"]
```
ameyakhare, line 29: I think you want to add typescript here for ts files

ameyakhare (changes requested):

.github/workflows/codeql.yml
joshma: AD in other PR, I think this is how you do typescript - let's merge this and I'll add typescript if GH support says we need it

ameyakhare (approved):

PR #39, opened 2022-06-13
Export remaining builtins tasks

ameyakhare:
Creates functions for interacting with all of the remaining builtins tasks.


email.ts
```diff
@@ -0,0 +1,31 @@
...
+    this.email = email;
+    this.name = name;
+  }
+}
+
+export default {
+  Sender,
+
+  message: async <Output = unknown>(
+    email_resource_id: string,
```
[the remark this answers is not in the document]
francescaleung: Did we decide on IDs to slugs?

rest.ts
```diff
@@ -0,0 +1,41 @@
...
+  BodyType,
+
+  request: async <Output = unknown>(
+    rest_resource_id: string,
+    method: Method,
+    path: string,
+    headers: Record<string, unknown> = {},
+    urlParams: Record<string, unknown> = {},
+    bodyType: BodyType = BodyType.Unknown,
+    body: unknown = null,
```
francescaleung, line 30: string or Record<string, unknown>

francescaleung (changes requested):
requesting changes for clarification on JSTs + slugs vs. ids

email.ts
```diff
@@ -0,0 +1,31 @@
...
+    sender: Sender,
+    recipients: unknown,
```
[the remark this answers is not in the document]
colinking: Let's not expose that. Most (all?) use-cases could be replaced with user code.

email.ts
```diff
@@ -0,0 +1,31 @@
...
+  constructor(email: string, name: string) {
+    this.email = email;
+    this.name = name;
+  }
+}
+
+export default {
+  Sender,
+
+  message: async <Output = unknown>(
```
colinking, line 16: @ameyakhare: wdyt of using objects instead of param lists? This solves the adjacent param type issue with `sql.query`.

```ts
email.message({
	resource: "..",
	sender: {...},
	// ...
})
```

rest.ts
```diff
@@ -0,0 +1,41 @@
...
+  Raw = "raw",
+  FormData = "form-data",
+  FormURLEncoded = "x-www-form-urlencoded",
+}
+
+export default {
+  Method,
+  BodyType,
+
+  request: async <Output = unknown>(
```
colinking, line 23: We know the return types for built-in tasks, so let's add that.

Let's also extract this out as a named/exported type:

```
export type RESTRequestOutput = {
	response: Record<string, unknown> | string
}
```

email.ts
colinking: Let's also expose a named/exported type for the inputs:

```ts
export type EmailMessageInput = {
	resource: string | Resource
	sender: EmailContact
  // ...
}
```

colinking (approved):

email.ts
```diff
@@ -0,0 +1,31 @@
+import { Run, ExecuteOptions, executeInternal } from "./tasks";
+
+export class Sender {
```
colinking, line 3: Prefer objects over classes

```ts
export type Sender = {
  email: string
  name: string
}
```

colinking (commented):
(whoops meant to just comment -- will defer to @francescaleung's review)

email.ts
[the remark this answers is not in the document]
ameyakhare: right now im using ids, and then in a follow up I'll be adding the translation code so that we start passing in slugs instead (so I can test that the SDK works without waiting for resource env var changes)

rest.ts
ameyakhare: this is a good idea, I'll try to find what the outputs look like

ameyakhare: @francescaleung I attempted to do this to the best of my ability

francescaleung (approved):
approving to unblock, would like to see tests in a future PR please :)

email.ts
```diff
@@ -1,30 +1,25 @@
...
+  email: string;
+  name: string;
+};
 
-  constructor(email: string, name: string) {
-    this.email = email;
-    this.name = name;
-  }
-}
+export type EmailRequestOutput = Record<string, number>;
```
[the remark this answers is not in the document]
colinking: It's a good question. Most modern `npm` packages are written to take advantage of tree-shaking where you directly import the function, rather than an object with a function. For types, you always import the type directly, so it's more often to prefix it.

`EmailRequestOutput` sgtm

mongodb.ts
```diff
@@ -0,0 +1,240 @@
+import { Run, ExecuteOptions, executeInternal } from "./tasks";
+
+export type DocumentRequestOutput = Record<string, unknown>;
+
+export type InsertOneRequestOutput = {
+  InsertedID: unknown;
```
colinking, line 6: If these are capitalized in our stdAPI outputs, then we're missing `json` tags... could you look into fixing that?

mongodb.ts
```diff
@@ -0,0 +1,240 @@
...
+
+export type InsertManyRequestOutput = {
+  InsertedIDs: unknown[];
+};
+
+export type UpdateRequestOutput = {
+  MatchedCount: number;
+  ModifiedCount: number;
+  UpsertedCount: number;
+  UpsertedID: unknown;
```
colinking, line 17: Hmm, is this not `string`? (I don't know the Mongo API)

email.ts
colinking: Okay. At the very least, we should only include required arguments as function parameters and collapse all optional arguments into a single object.

mongodb.ts
francescaleung: I think these are directly from the mongo library, pretty sure these aren't our structs.

I'd also worry about changing these now, in case folks are using mongodb outputs in runbooks.

email.ts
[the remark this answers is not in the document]
francescaleung: in this case I'd probably go with `EmailMessageOutput` (since the namespace is `email` & the function is `message`)

mongodb.ts
colinking: Okay, we'll want to convert those library structs into our own. In general, we should not be exporting third-party structs as part of our API. I created a Linear: https://linear.app/airplane/issue/AIR-4013/mongodb-built-in-output-is-pascal-case

rest.ts
```diff
@@ -16,25 +16,29 @@ export enum BodyType {
...
+  request: async <Output = RESTRequestOutput | undefined | null>(
+    restResourceID: string,
     method: Method,
     path: string,
     headers: Record<string, unknown> = {},
     urlParams: Record<string, unknown> = {},
     bodyType: BodyType = BodyType.Unknown,
-    body: unknown = null,
+    body: Record<string, unknown> | string | undefined | null = null,
     formData: Record<string, unknown> | undefined | null = null,
```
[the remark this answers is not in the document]
ameyakhare: I dont!

mongodb.ts
ameyakhare: 👍

mongodb.ts
ameyakhare: nope! its an interface{} in the api

ameyakhare: reading online it looks like its a hexadecimal string

PR #40, opened 2022-06-14
Modify exports and pass resources to durable execute

calufornia:
Contains various fixes related to utilizing the Node SDK in workflow tasks:
1. Exports all SDK methods as named exports, in addition to be included in the default export. The way that Temporal is bundling workflow code causes all exports (including the default export) to be wrapped in another (default) export object, i.e. if we do `import airplane from "airplane"`, the `airplane` object looks like:
```
{
  "namedExport": ...
  "default": {
    "execute": ...
  }
}
```
which means we have to write `airplane.default.execute(...)`. Exporting all SDK methods as named exports should allow us to retain the behavior of just calling `airplane.execute(...)` in the workflow task.
2. Moves `ExecuteOptions` to `api.ts` to remove a circular dependency `activities.ts` -> `tasks.ts` -> `activities.ts`.
3. Passes resources into the `durableExecute` method.
4. Renames `registerActivities` -> `_registerActivities`. We need to export this method since the worker shim needs to register the activities, but we make it more explicit that this is an internal method that is not meant to imported by users.

 
cc @colinking


ameyakhare (dismissed):
can you rebase things before I review

tasks.ts
```diff
@@ -30,16 +30,6 @@ export type Run<Input = unknown, Output = unknown> = {
...
 
-export type ExecuteOptions = {
-  host?: string;
-  token?: string;
-  apiKey?: string;
-  envID?: string;
-  envSlug?: string;
-  source?: string;
-  runtime?: string;
-};
```
colinking, line 41: +1 nice

index.ts
```diff
@@ -1,12 +1,14 @@
...
 import mongodb from "./mongodb";
 import { appendOutput, setOutput } from "./output";
 import rest from "./rest";
 import slack from "./slack";
 import sql from "./sql";
 import { execute } from "./tasks";
-export { ExecuteOptions } from "./tasks";
+export { ExecuteOptions } from "./api";
+
+export { appendOutput, setOutput, execute, sql };
```
colinking, line 11: TIL! Export the other stdapi functions too?

colinking (approved):

PR #41, opened 2022-06-15
Determine internal execute method using AsyncLocalStorage

calufornia:
Previously, we would have to include the runtime as part of the `ExecuteOptions` passed to `airplane.execute` in order to let the SDK know whether to call the internal execute method for standard tasks, or workflow tasks, i.e. define something like:
```
export default async function(params) {
  return await airplane.execute("rest_task", undefined, {
    runtime: "workflow",
  });
}
```
This is somewhat redundant since the code above is already being executed as a workflow task, and the user shouldn't have to specify again that it needs to run in a durable context.

We utilize the `AsyncLocalStorage` class provided by Temporal (which is basically just a context-aware wrapper around the one from Node's `async_hooks`), to pass in metadata (like the runtime) such that is globally accessible during Workflow execution. This way, we can set the runtime in async local storage in the workflow shim and fetch it during execution, so that we only need to define:
```
export default async function(params) {
  return await airplane.execute("rest_task");
}
```

Tested locally on both standard and workflow tasks.

### DIFF
--- a/async.ts
+++ b/async.ts
@@ -1,0 +1,6 @@
+import { AsyncLocalStorage } from "@temporalio/workflow";
+export type AirplaneStore = {
+  runtime?: string;
+};
+
+export const _storage = new AsyncLocalStorage<AirplaneStore>();
--- a/index.ts
+++ b/index.ts
@@ -5,6 +5,8 @@ import { execute } from "./tasks";
 export { ExecuteOptions } from "./tasks";
 export * as resources from "./resources";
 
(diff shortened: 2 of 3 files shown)


ameyakhare (approved):
nice now users dont have to specify its running within a workflow right

colinking (approved):
🔥

calufornia:
Closing this in favor of https://github.com/airplanedev/node-sdk/pull/41

PR #44, opened 2022-06-20
Refactor code layout

colinking:
This PR moves most of the codebase into an `internal` folder, which aims to discourage users from directly importing from it. Fields and types that are safe to import are exported from a file in the root. Updates the export format so that built-ins can be imported as `import { query } from 'airplane/sql'`.

cc @ameyakhare


colinking:
* main
  * **PR #44** <a href="https://app.graphite.dev/github/pr/airplanedev/node-sdk/44" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/airplanedev/node-sdk/44?utm_source=stack-comment).

internal/api/client.ts
```diff
@@ -0,0 +1,71 @@
...
+import { RunStatus } from "./types";
+
+export type ClientOptions = {
+  host?: string;
+  token?: string;
+  apiKey?: string;
+  envID?: string;
+  envSlug?: string;
+  source?: string;
+  runtime?: string;
```
calufornia, line 11: We can remove this now since we just read from `AIRPLANE_RUNTIME`

internal/api/client.ts
```diff
@@ -0,0 +1,71 @@
...
+
+  async getRunOutput<Output = unknown>(runID: string): Promise<Output> {
+    const { output } = await this.fetcher.get<{ output: Output }>("/v0/runs/getOutputs", {
+      id: runID,
+    });
+
+    return output;
+  }
+
+  async getRun<ParamValues = unknown>(runID: string) {
```
calufornia, line 59: Should we add a return type here?

internal/api/fetcher.ts
```diff
@@ -3,9 +3,9 @@ import withFetchRetries, { RequestInitWithRetry } from "fetch-retry";
 import type { FetchError } from "node-fetch";
 import querystring from "query-string";
 
-import { version } from "./package.json";
+import { version } from "../../package.json";
 
-export type FetchOptions = {
+export type FetcherOptions = {
```
calufornia, line 8: Maybe this could be:
```
ClientOptions & {
  retryDelay?: (attempt: number) => number;
}
```
unless we think the client/fetcher options will differ significantly

email.ts
```diff
@@ -1,26 +1,7 @@
-import { Run, ExecuteOptions, executeInternal } from "./tasks";
+// Supports:
+// import { message, MessageOutput } from 'airplane/email'
```
calufornia, line 2: 🔥

internal/runtime/index.ts
```diff
@@ -0,0 +1,22 @@
+import { ClientOptions } from "../api/client";
+import { Run } from "../api/types";
+import { runtime as standardRuntime } from "./standard";
+import { runtime as workflowRuntime } from "./standard";
```
calufornia, line 4: (Probably fixed by now, but just noting in case)

calufornia (approved):

internal/activities.ts
```diff
@@ -0,0 +1,24 @@
+import { Client, ClientOptions } from "./api/client";
+
+// Recommended activity factory by Temporal: https://docs.temporal.io/typescript/activities/#important-design-patterns
+//
+// We re-initialize the client in each activity because there seems to be an issue with Temporal serializing
+// classes - the object that is returned by the activity will have no methods associated with it.
+export const _registerActivities = () => ({
```
calufornia, line 7: Just confirming, since this still needs to be imported by the worker shim, is the proper way to do this now `import { _registerActivities from airplane/internal/activities`?

internal/api/client.ts
colinking: I'm being lazy and letting TS infer it since it isn't the `Run` type (it does not have output) 😛

PR #46, opened 2022-06-22
Replace `np` with automatic deployments from GHA

colinking:
We were having issues with our previous release process that used `np`. First, it depended on engineers pushing a commit to `main` to change the version field in `package.json`. This does not work with our branch protection rules. Second, it does not support deploying a sub-directory. We build our TS into the `dist` folder and want the built code to be the root of the released package (so that users can import `airplane/sql` which depends on a `sql.js` file in the root of the package). There is a `contents` field from `np` that supports this , but I couldn't get this to work the way we needed.

This PR moves the release process to GHA. It's much more convenient -- just merge a PR that bumps the version in the `package.json` and it will automatically publish. I tested this by releasing `v0.2.0-10`.

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -39,7 +39,7 @@ jobs:
           echo "Version: ${VERSION}"
 
           if [ ! "$(npm show airplane@${VERSION} version)" = "" ]; then
-            echo "Version is already published. Skipping."
+            echo "Version is already published, skipping: https://www.npmjs.com/package/airplane/v/${VERSION}"
             exit 0
           fi
 
@@ -62,7 +62,8 @@ jobs:
           TMP_DIR=$(mktemp -d -t publish-XXXXXXXXXX)
           TARBALL="${TMP_DIR}/package.tgz"
           yarn pack -f ${TARBALL}
-          tar -tf ${TARBALL} | tree --fromfile .
+          tar -xf ${TARBALL} -C ${TMP_DIR}
+          tree ${TMP_DIR}/package
 
           echo "yarn publish ${TARBALL} --new-version ${VERSION} --tag=${TAG}"
           # TODO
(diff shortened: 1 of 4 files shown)


.github/workflows/tests.yml
```diff
@@ -23,3 +23,60 @@ jobs:
...
+          TARBALL="${TMP_DIR}/package.tgz"
+          yarn pack -f ${TARBALL}
+          tar -xf ${TARBALL} -C ${TMP_DIR}
+          tree ${TMP_DIR}/package
+
+          yarn publish ${TARBALL} --new-version ${VERSION} --tag=${TAG}
+          echo "::endgroup::"
+
+          echo "::group::Creating GitHub release..."
+          gh release create ${VERSION} --generate-notes ${GH_ARGS} --target $(git rev-parse HEAD)
```
calufornia, line 80:
```suggestion
          gh release create "v${VERSION}" --generate-notes ${GH_ARGS} --target $(git rev-parse HEAD)
```

calufornia: Or are we no longer prepending releases with `v`?

calufornia (approved):
GHA job lgtm! my main suggestion would be to put it in a separate workflow file? Seems a bit off to have it in `tests.yml` and just skipped in all PR commits.

.github/workflows/tests.yml
colinking: Good catch! The NPM version is not prefixed, but the git tag + release should be. Thanks!

colinking:
> GHA job lgtm! my main suggestion would be to put it in a separate workflow file? Seems a bit off to have it in `tests.yml` and just skipped in all PR commits.

Agreed. We just need this to wait until the tests pass, which requires it to be in the same workflow. One workaround would be to move this to a separate workflow that is invoked by a job in the `tests` workflow? Bit more complicated. Leaving as-is for now.

PR #47, opened 2022-06-23
Remove error if no authn is provided

leeweisberger:
In production, views rely on session cookies to authn with the API - they do not explicitly pass in an authn token or api key. This error precludes this from working.


colinking (approved):

colinking:
FYI, there's some in-flight changes to the SDK release process in this PR: https://github.com/airplanedev/node-sdk/pull/46

PR #57, opened 2022-06-28
Add basic prompts: `airplane.input.text` 

colinking:
Adds support for a single type of prompt: text strings. Follow-up PRs will add the remainder of the prompt kinds.

cc @kellymtrinh 

Resolves AIR-4114

### DIFF
--- a/internal/api/client.ts
+++ b/internal/api/client.ts
@@ -1,5 +1,5 @@
 import { Fetcher } from "./fetcher";
-import { RunStatus } from "./types";
+import { ParamSchema, Prompt, RunStatus } from "./types";
 
 export type ClientOptions = {
   host?: string;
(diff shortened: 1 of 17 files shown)


internal/input/slug.ts
```diff
@@ -0,0 +1,37 @@
+const maxLength = 50;
+
+// makeSlug will convert a string into a valid slug, according to the
+// semantics of slugs used in the Airplane API. Emulates the
+// same gosimple/slug-based implementation.
+export const makeSlug = (s: string): string => {
```
colinking, line 6: Ported over from `airplanedev/web`. Would be nice to share as a package at some point, but just duplicating for now.

internal/api/client.ts
```diff
@@ -47,20 +47,34 @@ export class Client {
...
-      paramValues: ParamValues;
+      paramValues: P;
       taskID: string;
     }>("/v0/runs/get", { id: runID });
   }
+
+  async createPrompt(params: ParamSchema[]): Promise<string> {
+    const resp = await this.fetcher.post<{
+      id: string;
+    }>("/v0/prompts/create", { schema: { parameters: params }, values: null });
```
francescaleung, line 70: is it weird that this needs to send values?

francescaleung (approved):

internal/api/client.ts
colinking: Ah, it doesn't need to. The idea is you can submit "default values". We don't use that feature yet, though. Will remove from the client here.

colinking:
@francescaleung I'm also bumping the version to `-12`

PR #59, opened 2022-06-30
Use resource aliases in builtin functions

ameyakhare:
Resolves AIR-4143

Switches over from resource ids to aliases now that resource attachments are out. Users will need to attach the resource and then refer to the resource via alias in all builtin functions.


internal/builtins/builtins.ts
```diff
@@ -0,0 +1,18 @@
+export const convertResourceAliasToID = (alias: string) => {
+  const resources = JSON.parse(process.env.AIRPLANE_RESOURCES ?? "{}");
+  const resource_version = process.env.AIRPLANE_RESOURCES_VERSION ?? null;
```
francescaleung, line 3:
```suggestion
  const resourceVersion = process.env.AIRPLANE_RESOURCES_VERSION ?? null;
```

francescaleung (approved):

PR #62, opened 2022-07-01
Update Built-ins optional parameters to be an obj

kellymtrinh:
**For our built-ins library, updates the function parameters so optional arguments are in an object parameter:**

Before:
```
rest.request("resourceID",   rest.Method.Get, "/testelement", null, {page: 1},  null, null, null, {something})
```
After:
```
rest.request("resourceID",  rest.Method.Get, "/testelement", {urlParams: {page: 1}})
```

- If a param was optional or we were setting the default value to null (because it's optional in the underlying API) I moved it to an obj. 
- Combined the client options and input options into one "opts" object to match what we did here: https://github.com/airplanedev/node-sdk/blob/ed5b38cbdbcab52a79909664519eeeea402ea783/internal/input/index.ts#L6

CC: @ameyakhare


internal/builtins/mongodb.ts
```diff
@@ -8,13 +8,16 @@ export type DocumentOutput = Record<string, unknown>;
...
-  skip: number | undefined | null = null,
-  limit: number | undefined | null = null,
+  params?: {
+    filter: Record<string, unknown> | undefined | null;
+    projection: Record<string, unknown> | undefined | null;
+    sort: Record<string, unknown> | undefined | null;
+    skip: number | undefined | null;
+    limit: number | undefined | null;
+  },
   opts?: ClientOptions
```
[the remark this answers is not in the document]
colinking: If we merge them, let's call them options to align with "InputOptions" and "ClientOptions".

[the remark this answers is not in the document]
kellymtrinh: Merging them into one obj: `opts: InputOptions` which has a client field of type ClientOptions

internal/builtins/mongodb.ts
```diff
@@ -8,85 +8,94 @@ export type DocumentOutput = Record<string, unknown>;
...
   return getRuntime().execute(
     "airplane:mongodb_find",
     { collection, filter, projection, sort, skip, limit },
     { db: convertResourceAliasToID(mongodbResource) },
-    opts
+    client
   );
 };
 
+export type InputOptions = {
```
colinking, line 29: We called it `InputOptions` because it's options for inputs/prompts. Can we make this more specific? Looks like this is used for the `find*` methods -- how about `FindOptions`

internal/builtins/rest.ts
```diff
@@ -22,21 +22,26 @@ export type RequestOutput = {
   response: Record<string, unknown> | string;
 };
 
+export type InputOptions = {
+  headers?: Record<string, unknown>;
```
colinking, line 26: If we accept `null` in some places for optional fields, let's accept it everywhere.

```suggestion
  headers?: Record<string, unknown> | null;
```

colinking (changes requested):
LGTM. Can you update `email.message` too? (pull subject/message into `opts`)

For consistency, let's also move `client` into an options object for all built-ins, e.g. `slack.message`. Otherwise, it's a breaking change to switch to an options object.

internal/builtins/rest.ts
kellymtrinh: Headers and urlParams need to be `{}` and not null or undefined :/ - so they're not actually optional but we do have a default value.

```
let test = {"a" : null}
let {a = "what"} = test // a is null and not what
let {none = "what2"} = test // none is what2
```
I think letting headers be null means we have to add an extra check for if it's null vs. only handling if it's undefined or not.

colinking: I'm referring to what we accept as input, which can differ from what we pass to the API. Can we just convert both `null` and `undefined` to `{}`? `headers = headers ?? {}`

internal/builtins/email.ts
```diff
@@ -8,20 +8,25 @@ export type Contact = {
...
 export const message = async (
   emailResource: string,
   sender: Contact,
   recipients: Contact[] | string[],
-  subject = "",
-  message = "",
-  opts?: ClientOptions
+  opts: EmailOptions = {}
 ): Promise<Run<ParamValues, MessageOutput>> => {
+  const { subject = "", message = "", client } = opts;
```
colinking, line 25: Just checked, and this only applies the default value if `undefined` _not_ if `null`:

```ts
let { foo = "" } = {}; foo
''
let { foo = "" } = {foo: null}; foo
null
let { foo = "" } = {foo: undefined}; foo
''
```

I'm assuming we don't want to differentiate between the two? And convert both `null/undefined` to the default value.

internal/builtins/mongodb.ts
```diff
@@ -8,85 +8,96 @@ export type DocumentOutput = Record<string, unknown>;
...
-  limit: number | undefined | null = null,
-  opts?: ClientOptions
+  opts: {
+    filter?: Record<string, unknown> | null;
+    projection?: Record<string, unknown> | null;
+    sort?: Record<string, unknown> | null;
+    skip?: number | null;
+    limit?: number | null;
+    client?: ClientOptions;
+  } = {}
```
colinking, line 18:
```suggestion
  opts: FindOptions & {
    skip?: number | null;
    limit?: number | null;
  } = {}
```

internal/builtins/mongodb.ts
```diff
@@ -8,85 +8,96 @@ export type DocumentOutput = Record<string, unknown>;
...
   collection: string,
-  filter: Record<string, unknown> | undefined | null = null,
-  projection: Record<string, unknown> | undefined | null = null,
-  sort: Record<string, unknown> | undefined | null = null,
-  opts?: ClientOptions
+  opts: FindOptions = {}
 ): Promise<Run<ParamValues, DocumentOutput | undefined | null>> => {
+  const { filter, projection, sort, client } = opts;
   return getRuntime().execute(
     "airplane:mongodb_findOne",
```
colinking, line 49: Hmm, unrelated to your change, but this looks like a bug -- capital letters are not allowed as slugs. @ameyakhare can you TAL?

```suggestion
    "airplane:mongodb_findone",
```

internal/builtins/mongodb.ts
```diff
@@ -8,85 +8,96 @@ export type DocumentOutput = Record<string, unknown>;
...
   collection: string,
   replacement: Record<string, unknown>,
-  filter: Record<string, unknown> | undefined | null = null,
-  projection: Record<string, unknown> | undefined | null = null,
-  sort: Record<string, unknown> | undefined | null = null,
-  upsert: boolean | undefined | null = null,
-  opts?: ClientOptions
+  opts: FindOptions & {
+    upsert?: boolean | null;
+  } = {}
```
colinking, line 92:
```suggestion
  opts: FindOptions & UpdateOptions = {}
```

internal/builtins/mongodb.ts
```diff
@@ -8,85 +8,96 @@ export type DocumentOutput = Record<string, unknown>;
...
   return getRuntime().execute(
     "airplane:mongodb_find",
     { collection, filter, projection, sort, skip, limit },
     { db: convertResourceAliasToID(mongodbResource) },
-    opts
+    client
   );
 };
 
+export type FindOptions = {
```
colinking, line 29: Sorry, just noticed `find` actually has more options. This name might be confusing then. How about `QueryOptions`?  MongoDB uses query/update in their docs: https://www.mongodb.com/docs/manual/tutorial/query-documents/

internal/builtins/rest.ts
```diff
@@ -22,21 +22,26 @@ export type RequestOutput = {
   response: Record<string, unknown> | string;
 };
 
+export type RequestOptions = {
+  headers?: Record<string, unknown>;
+  urlParams?: Record<string, unknown>;
+  bodyType?: BodyType | undefined | null;
```
colinking, line 28:
```suggestion
  bodyType?: BodyType | null;
```

internal/builtins/sql.ts
```diff
@@ -12,17 +12,22 @@ export enum TransactionMode {
 
 export type QueryOutput = Record<string, unknown[]>;
 
+export type QueryOptions = {
+  args?: Record<string, unknown> | null;
+  transactionMode?: TransactionMode;
```
colinking, line 17:
```suggestion
  transactionMode?: TransactionMode | null;
```

colinking (commented):
Don't forget `slack.message`!

colinking (approved):
Otherwise, LGTM. Thanks for working through my feedback!

Can you bump the version, too? Just bump it in `package.json` and it'll get auto-released from `main`.

internal/builtins/mongodb.ts
ameyakhare: actually this is correct, builtins are special cased since the builtin namespace is `mongodb` and the function is `findOne` (capital o)

colinking: Hmm, ideally this is a valid slug. Looks like only MongoDB has non-slug names. What if we add "aliases" here for each of those? https://github.com/airplanedev/airport/blob/8573361961ef526398947b834ae0e3493484a603/pkg/stdapi/mongodb/mongodb.go#L10-L24

e.g.

```go
		"findOneAndUpdate":  func() function.FunctionHandler { return handlers.FindOneAndUpdateHandler{} },
		"find_one_and_update":  func() function.FunctionHandler { return handlers.FindOneAndUpdateHandler{} },
```

internal/builtins/email.ts
kellymtrinh: Oof yeah that's the same case in the other one with rest headers and urlParams that I fixed- I feel like optional params should be undefined or a value, and we shouldn't let users set it to be null?? (It would be null if explicitly set?) Users should either set a good value, or not set it. But they shouldn't set it as null? 

I will update what I did in rest headers though and destructure as one step, and correctly set it outside of that using 
`  subject: subject ?? ""`

colinking: I'd prefer we be liberal in what we accept in our SDKs. For example, if we didn't allow `null` then this code wouldn't work unless you used `let headers = undefined`. If we're going to accept `undefined` might as well also accept `null`.

```ts
let headers = null
if (useAuth) {
  headers = {"Authorization": getAuth()}
}
airplane.rest.request(..., {headers})
```

PR #67, opened 2022-07-06
Add support for setting output in workflow tasks

calufornia:
This PR adds workflow task support for the `setOutput` and `appendOutput` methods of our SDK - we move `logChunks` to the runtime interface because its implementation varies depending on whether the task is a standard or workflow task. 

For standard tasks, we log to stdout and utilize the uuid package to generate chunk keys. For workflow tasks, however, we must log to a sink , and also use Temporal's `uuid` method, which generates UUIDs with a seed that is derived from the workflow id, and so guarantees workflow determinism.

Also changes any references of `substr` -> `substring` since the former has been deprecated.

### DIFF
--- a/internal/output.ts
+++ b/internal/output.ts
@@ -1,4 +1,4 @@
-import { v4 as uuidv4 } from "uuid";
+import { getRuntime } from "./runtime";
 
 /**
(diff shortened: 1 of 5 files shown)


colinking (approved):
🔥 

Make sure to bump the version in `package.json`, too.

PR #68, opened 2022-07-06
Throw error when child runs fail

colinking:
When a run fails (or is cancelled), throw an error that a user can optionally catch. IOW, by default a failed child run will fail the parent run.

Resolves AIR-4093

### DIFF
--- a/internal/execute.ts
+++ b/internal/execute.ts
@@ -1,5 +1,6 @@
 import { ClientOptions } from "./api/client";
-import { ParamValues, Run } from "./api/types";
+import { ParamValues, Run, RunStatus } from "./api/types";
+import { RunError } from "./errors";
 import { getRuntime } from "./runtime";
 
 export const execute = async <Output = unknown>(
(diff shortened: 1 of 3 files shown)


colinking:
* main
  * **PR #68** <a href="https://app.graphite.dev/github/pr/airplanedev/node-sdk/68" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/airplanedev/node-sdk/68?utm_source=stack-comment).

internal/execute.ts
```diff
@@ -1,11 +1,18 @@
...
   slug: string,
   params: ParamValues = {},
   opts: ClientOptions = {}
 ): Promise<Run<typeof params, Output>> => {
-  return getRuntime().execute(slug, params, {}, opts);
+  const run = await getRuntime().execute<Output>(slug, params, {}, opts);
+
+  if (run.status === RunStatus.Failed || run.status === RunStatus.Cancelled) {
+    throw new RunTerminationError(run);
+  }
```
leeweisberger, line 15: Don't want to slow you guys down, but just calling out that many of the node-sdk releases have included breaking changes that we have to keep fixing in the views code. Would be helpful to call these out in release notes / bump our lib version more aggressively.

We're also missing types that are kind of necessary to use this library - e.g. as a consumer of this package I need the RunTerminationError type, but it's not exported at the top level. Same with all of the other types used in top level, public functions (ParamValues, etc...)

PR #75, opened 2022-07-18
Import nested temporal package

leeweisberger:
We are getting warning on dep install that we should be using nested temporal packages:

image

Switch to using the nested workflow package. 

I'm unsure how to manually test this - any guidance here?

Unrelated, but we need to work on the build system for this repo. It doesn't produce esm modules so it isn't tree shakeable - this is causing consumers (like views) to import all of the code even though we are only using part of the code (e.g. views imports temporal).

### DIFF
--- a/package.json
+++ b/package.json
@@ -48,10 +48,10 @@
     "printWidth": 100
   },
   "dependencies": {
+    "@temporalio/workflow": "1.0.0-rc.0",
     "cross-fetch": "3.1.5",
     "fetch-retry": "5.0.3",
     "query-string": "7.1.1",
-    "temporalio": "1.0.0-rc.0",
     "uuid": "8.3.2"
   }
 }
--- a/package.json
+++ b/package.json
@@ -48,6 +48,7 @@
     "printWidth": 100
   },
   "dependencies": {
+    "@temporalio/worker": "1.0.0-rc.1",
     "@temporalio/workflow": "1.0.0-rc.0",
     "cross-fetch": "3.1.5",
     "fetch-retry": "5.0.3",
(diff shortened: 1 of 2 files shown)


calufornia (approved):
The node shim also [requires @temporalio/worker](https://github.com/airplanedev/lib/blob/a45ef8c7b1b71b24d2383c38f5e30efc58f4fb22/pkg/build/workflow/worker-and-activity-shim.js#L1) and currently gets it (transitively I think) from the [node SDK](https://github.com/airplanedev/lib/blob/a45ef8c7b1b71b24d2383c38f5e30efc58f4fb22/pkg/build/node.go#L273), so we'll just need to make a small change to lib to install `@temporalio/worker`, I can make a follow-up PR.

calufornia:
> The node shim also [requires @temporalio/worker](https://github.com/airplanedev/lib/blob/a45ef8c7b1b71b24d2383c38f5e30efc58f4fb22/pkg/build/workflow/worker-and-activity-shim.js#L1) and currently gets it (transitively I think) from the [node SDK](https://github.com/airplanedev/lib/blob/a45ef8c7b1b71b24d2383c38f5e30efc58f4fb22/pkg/build/node.go#L273), so we'll just need to make a small change to lib to install `@temporalio/worker`, I can make a follow-up PR.

Actually, I take that back - we might need to install `@temporalio/worker` here as well. There's currently a bug where if temporalio is included in the dependencies of a user's code, and additionally in the node shim, the temporal workflow won't progress (I've tested this locally). We should eventually figure out why this happens so that we don't need to depend on the node SDK for temporal dependencies, but for now I think we should also just install the worker package.

PR #76, opened 2022-07-21
Add node configuration as code types to sdk

ameyakhare:
Resolves AIR-4314
Resolves AIR-4315

Adds the type system for code-only configuration, including a default airplane.Task export that denotes the function to wrap a user task.


internal/codeConfiguration/paramTypes.ts
```diff
@@ -0,0 +1,115 @@
...
+  datetime: string;
+  configvar: string;
+};
+
+export type JSParamValues = {
+  shorttext: string;
+  longtext: string;
+  sql: string;
+  boolean: boolean;
+  upload: string;
```
colinking, line 48: Uploads are serialized as objects not strings.

```
type File = {
  __airplaneType: "upload"
  id: string
	url: string
}
```

We also call them files externally, so let's use "file" wherever we can.

internal/codeConfiguration/task.ts
```diff
@@ -0,0 +1,48 @@
...
+  parameters?: TParams;
+  requireRequests?: boolean;
+  allowSelfApprovals?: boolean;
+  timeout?: number;
+  constraints?: Record<string, string>;
+  runtime?: "standard" | "workflow";
+};
+
+// eslint-disable-next-line  @typescript-eslint/no-explicit-any
+export type UserFunc<TParams extends Params> = (params: ParamValues<TParams>) => any;
```
colinking, line 23: Add a generic for the output so the output type is retained when a user calls a task function directly:

```suggestion
export type UserFunc<TParams extends Params, TOutput> = (params: ParamValues<TParams>) => TOutput;
```

internal/codeConfiguration/task.ts
```diff
@@ -0,0 +1,48 @@
...
+
+export type AirplaneFunc<TParams extends Params> = (
+  params: ParamValues<TParams>
+) => Promise<Awaited<ReturnType<UserFunc<TParams>>>>;
+
+export const Task = <TParams extends Params>(
+  config: TaskConfig<TParams>,
+  f: UserFunc<TParams>
+): AirplaneFunc<TParams> => {
+  const inAirplaneRuntime = !!(process.env.AIRPLANE_RUN_ID ?? "");
```
colinking, line 33: Let's use `AIRPLANE_RUNTIME` here.

internal/codeConfiguration/task.ts
```diff
@@ -0,0 +1,48 @@
...
+  constraints?: Record<string, string>;
+  runtime?: "standard" | "workflow";
+};
+
+// eslint-disable-next-line  @typescript-eslint/no-explicit-any
+export type UserFunc<TParams extends Params> = (params: ParamValues<TParams>) => any;
+
+export type AirplaneFunc<TParams extends Params> = (
+  params: ParamValues<TParams>
+) => Promise<Awaited<ReturnType<UserFunc<TParams>>>>;
```
colinking, line 27: What's this for? `ReturnType<UserFunc<TParams>>` is always `any` right?

colinking: Is this supposed to define the `config/baseFunc` fields? Not sure how it is different from UserFunc

internal/codeConfiguration/paramTypes.ts
```diff
@@ -0,0 +1,115 @@
+// Attributes for different types of params
```
colinking, line 1: Organization-wise, can you move the parameter-related logic into `internal/parameters.ts`? I'll likely end up using this logic for prompts, too.

The logic that is config-related (e.g. `ParamSchemaKind`) can then live in `internal/config.ts`.

internal/codeConfiguration/task.ts
```diff
@@ -0,0 +1,48 @@
...
+
+export const Task = <TParams extends Params>(
```
colinking, line 29: JS convention is that functions are lowercase and classes are uppercase. Let's lowercase here:

```suggestion
export const task = <TParams extends Params>(
```

internal/codeConfiguration/paramTypes.ts
```diff
@@ -0,0 +1,115 @@
...
+  kind: "datetime";
+};
+
+export type ConfigVarParam = SelectableParam<JSParamTypes["configvar"]> & {
+  kind: "configvar";
+};
+
+// Used by task templating
+
+export type ParamSchema =
```
colinking, line 99: Minor, keep it consistent with `ShortTextParam`. Alternatively, rename `ShortTextParam` -> `ShortTextParamSchema`.

```suggestion
export type Param =
```

internal/codeConfiguration/task.ts
```diff
@@ -0,0 +1,48 @@
+import { execute } from "../execute";
+import { ParamSchema, JSParamValues, ParamKinds, ParamSchemaKind } from "./paramTypes";
+
+export type Params = Record<string, ParamKinds | ParamSchema>;
+
+export type ParamValues<PSchema extends Params> = {
```
colinking, line 6: Let's use `T<...>` as a convention for types, to avoid naming collisions:

```suggestion
export type ParamValues<TParams extends Params> = {
```

colinking (changes requested):

internal/codeConfiguration/task.ts
colinking: > problem is a valid value for AIRPLANE_RUNTIME is empty, wouldn't this fail?

Pulling over from: https://github.com/airplanedev/cli/pull/520#discussion_r927076314

It should be set to either `standard | workflow | local`.

internal/codeConfiguration/task.ts
ameyakhare: this is the promised/awaited return type since running an airplane task is async. users can technically write a sync function, but we still need the return type to be a promise (async)

internal/config.ts
```diff
@@ -0,0 +1,52 @@
+import { execute } from "./execute";
+import { Param, JSParamValues, ParamKinds } from "./parameters";
+import { RuntimeKind } from "./runtime";
+
+export type ParamKind<PSchema extends ParamKinds | Param> = PSchema extends Param
```
colinking, line 5:
```suggestion
export type ParamKind<TParam extends ParamKinds | Param> = TParam extends Param
```

internal/config.ts
```diff
@@ -0,0 +1,52 @@
...
+  ? PSchema["kind"]
+  : PSchema;
+
+export type Params = Record<string, ParamKinds | Param>;
+
+export type ParamValues<TParams extends Params> = {
+  [PSlug in keyof TParams]: JSParamValues[ParamKind<TParams[PSlug]>];
```
colinking, line 12: The `T<...>`convention just helps skim a type and pick out which types are generics and which aren't:

```suggestion
  [TSlug in keyof TParams]: JSParamValues[ParamKind<TParams[TSlug]>];
```

internal/config.ts
```diff
@@ -0,0 +1,52 @@
...
+  description?: string;
+  parameters?: TParams;
+  requireRequests?: boolean;
+  allowSelfApprovals?: boolean;
+  timeout?: number;
+  constraints?: Record<string, string>;
+  runtime?: "standard" | "workflow";
+};
+
+// eslint-disable-next-line  @typescript-eslint/no-explicit-any
```
colinking, line 27:
```suggestion
```

internal/config.ts
```diff
@@ -0,0 +1,52 @@
...
+export const task = <TParams extends Params, TOutput>(
+  config: TaskConfig<TParams>,
+  f: UserFunc<TParams, TOutput>
+): AirplaneFunc<TParams, TOutput> => {
+  const runtime = process.env.AIRPLANE_RUNTIME ?? "";
+  const inAirplaneRuntime = runtime === RuntimeKind.Standard || runtime === RuntimeKind.Workflow;
+
+  const wrappedF = async (params: ParamValues<TParams>): Promise<Awaited<TOutput>> => {
+    if (inAirplaneRuntime) {
+      return (await execute(config.slug, params)).output as Awaited<TOutput>;
```
colinking, line 43:
```suggestion
      return (await execute<Awaited<TOutput>>(config.slug, params)).output;
```

internal/config.ts
```diff
@@ -0,0 +1,52 @@
...
+    }
+    return f.apply(null, [params]) as Awaited<TOutput>;
```
colinking, line 45:
```suggestion
    return await f.apply(null, [params]);
```

internal/parameters.ts
```diff
@@ -0,0 +1,117 @@
...
+  | SQLParam
+  | BooleanParam
+  | FileParam
+  | IntegerParam
+  | FloatParam
+  | DateParam
+  | DatetimeParam
+  | ConfigVarParam;
+
+export type ParamKinds = Param["kind"];
```
colinking, line 117:
```suggestion
export type ParamKind = Param["kind"];
```

internal/parameters.ts
```diff
@@ -0,0 +1,117 @@
...
+  longtext: string;
+  sql: string;
+  boolean: boolean;
+  file: File;
+  integer: number;
+  float: number;
+  date: string;
+  datetime: string;
+  configvar: ConfigVar;
+};
```
colinking, line 61: Why both? It this to support how config vars can be referenced by slug rather than full `__airplaneType`? Might be a naming thing.

internal/config.ts
```diff
@@ -0,0 +1,52 @@
...
+
+export type Params = Record<string, ParamKinds | Param>;
+
+export type ParamValues<TParams extends Params> = {
+  [PSlug in keyof TParams]: JSParamValues[ParamKind<TParams[PSlug]>];
+};
+
+export type TaskConfig<TParams extends Params> = {
+  slug: string;
+  name: string;
```
colinking, line 17: Can name be optional? And default to `slug`. Nice for the "microtask" use-case.

internal/parameters.ts
```diff
@@ -0,0 +1,117 @@
+// Attributes for different types of params
+export type ParamBase<T> = {
+  kind: string;
```
colinking, line 3: Just noticed this is `type` in our definition file. Should we keep it consistent with that?

internal/parameters.ts
```diff
@@ -0,0 +1,117 @@
...
+  name?: string;
+  description?: string;
+  optional?: boolean;
```
colinking, line 6: This is also `required` in the task definition.

colinking (approved):

internal/parameters.ts
ameyakhare: param types are for the configuration, and param values are what they show up as in the params object passed in to the user function

PR #77, opened 2022-07-22
Fix airplane runtime detection

ameyakhare:
The standard runtime can set the value of the runtime variable to a blank string. Check instead that the runtime env var is defined and not "dev".


colinking (changes requested):
AD: let's use AIRPLANE_RUNTIME

colinking (approved):

PR #78, opened 2022-07-26
Remove reliance on process

leeweisberger:
Remove reliance on process so this can be used outside of node.

If called from outside of node, always execute in the airplane runtime. We might revisit this in the future, but I think this is the correct path at least for now.


ameyakhare (changes requested):
as discussed offline: this will run the function natively when in a view context, AIRPLANE_RUNTIME needs to be set to some value when this code is run within a view

PR #81, opened 2022-07-28
Always include source when calling airplane API

calufornia:
Always include a source in the headers when making a request to the airplane api - if one isn't explicitly specified, we use `"sdk/node"`


leeweisberger (commented):
What's the usecase here?

calufornia:
> What's the usecase here?

This is for fixing the missing run source issue for child runs!

PR #83, opened 2022-07-29
Export File Type for Node task development

ianfabs:
Using airplane at work, it would be really nice if this type was available from the airplane package by default. I often use node scripts to process excel spreadsheets and it took me too long to figure out that this was the issue. I appreicate the airplane dev teams time and attention to this issue

### DIFF
--- a/index.ts
+++ b/index.ts
@@ -34,3 +34,4 @@ export default {
 // Exported types
 export type { ClientOptions } from "./internal/api/client";
 export type { RunStatus, Run } from "./internal/api/types";
+export type { File } from "./internal/parameters";


colinking:
Hey @ianfabs! We've hit this issue internally, too. We are planning to introduce a new `File` concept for the node SDK that will make it easier to work with files:

```ts
export class File extends Blob {
  public __airplaneType = "upload";
  public readonly id: string;
  public readonly url: string;

  constructor(source: string | Blob) {
    // ...
  }
}

// You can initialize a file from a URL or with file system APIs:
const file = new File("https://...")
const file = new File(new Blob([...]))
const file = new File(new Blob([await fs.readFile("./fools.json")]))

// By building on Blobs, users can decide when/how to fetch the file contents.
const string: string = await file.text()
const binary: ArrayBuffer = await file.arrayBuffer()
const stream: ReadableStream = file.stream()
```

This type will be exported and automatically applied to parameter types when using the (soon to be announced) [config-as-code](https://github.com/airplanedev/node-sdk/blob/05cafc11e4ae08b9dcf6a3965c8b69b53e5efb5d/examples/workflow/main.airplane.ts#L4).

I'm going to close for now, but I'll follow-up here when the new `File` support lands.

colinking:
* main
  * **PR #83** <a href="https://app.graphite.dev/github/pr/airplanedev/node-sdk/83" target="_blank">Graphite</a>  👈
    * **PR #190** <a href="https://app.graphite.dev/github/pr/airplanedev/node-sdk/190" target="_blank">Graphite</a> 

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/airplanedev/node-sdk/83?utm_source=stack-comment).

colinking:
Hey @ianfabs -- we released the new `AirplaneFile` type in `0.2.26` which abstracts away the logic for downloading a file's contents.

If you're using [inline config](https://docs.airplane.dev/tasks/js-inline-config) to specify your tasks, your parameters will be automatically typed (including files!).

PR #87, opened 2022-08-02
Extract workflow-specific logic into @airplane/workflow-runtime

calufornia:
This PR moves workflow-specific implementations of the `RuntimeInterface` into a separate `@airplane/workflow-runtime` package, primarily because the temporal dependencies are large and take a long time to install due to a Rust build of temporal's core. We use yarn workspaces to help manage dependencies between `@airplane/workflow-runtime` and the base `airplane` package.

We maintain an internal `RuntimeMap` that maps a runtime kind to a runtime interface, which is initialized with just the dev and standard runtimes. Separately, for workflow tasks, the node shim will install `@airplane/workflow-runtime` and call `registerWorkflowRuntime()` to include the workflow `RuntimeInterface` in the runtime map.

We set up the the packages in the following directory structure - having a top-level `packages` directory with all packages in the monorepo seems to be standard (e.g. https://github.com/vercel/vercel/tree/main/packages):
```
- packages/
  - airplane/
    - package.json
  - workflow-runtime/
    - package.json
```

We try to keep most of the dev dependencies in the root package.json, and prefer linting/testing/etc. at the root level - otherwise, we need to install the dev dependency packages inside each package.


packages/workflow-runtime/package.json
```diff
@@ -0,0 +1,22 @@
...
+  },
+  "engines": {
+    "node": "^14 || ^16 || ^18"
+  },
+  "dependencies": {
+    "@temporalio/worker": "1.0.0",
+    "@temporalio/workflow": "1.0.0",
+    "airplane": "0.2.0-22"
+  },
+  "airplane": {
```
calufornia, line 19: This is needed for dev-ing Node SDK changes for workflow tasks, since workspaces is enabled.

packages/workflow-runtime/tsconfig.json
```diff
@@ -0,0 +1,14 @@
+{
+  "compilerOptions": {
+    "lib": ["es2019", "es2020.promise", "es2020.bigint", "es2020.string"],
+    "module": "commonjs",
+    "target": "es2019",
+    "strict": true,
+    "esModuleInterop": true,
+    "skipLibCheck": true,
+    "forceConsistentCasingInFileNames": true,
+    "outDir": "dist/internal",
```
calufornia, line 10: I noticed that running `yarn build` for the workflow-runtime causes the `internal` folder to get removed and instead its contents to get pushed up, i.e. instead of
```
- dist/
  - internal/
    - index.d.ts
    - index.js
```
we would get 
```
- dist/
  - index.d.ts
  - index.js
```
This might be because there's no "top-level" `index.ts` - to include the `internal` folder (which is preferable so users don't attempt to import methods from this package), we specify a different `outDir`

packages/airplane/package.json
```diff
@@ -0,0 +1,22 @@
...
+  "name": "airplane",
+  "version": "0.2.0-22",
+  "description": "Node.js SDK for writing Airplane.dev tasks",
+  "main": "index.js",
+  "types": "index.d.ts",
+  "repository": "airplanedev/node-sdk",
+  "author": "hello@airplane.dev",
+  "license": "MIT",
+  "scripts": {
+    "build": "rm -rf dist && mkdir dist && cp ../../LICENSE ../../README.md package.json dist && tsc -p tsconfig.build.json"
```
calufornia, line 11: Removed the `-t` flag since MacOS doesn't come with gnu coreutils by default

packages/workflow-runtime/package.json
```diff
@@ -0,0 +1,22 @@
...
+  "scripts": {
+    "build": "rm -rf dist && mkdir -p dist/internal && cp ../../LICENSE ../../README.md package.json dist && tsc"
+  },
+  "engines": {
+    "node": "^14 || ^16 || ^18"
+  },
+  "dependencies": {
+    "@temporalio/worker": "1.0.0",
+    "@temporalio/workflow": "1.0.0",
+    "airplane": "0.2.0-22"
```
calufornia, line 17: I wasn't able to get `"airplane": "*"` working. A `yarn install` in the root would install the version of package tagged with `latest` (i.e. 0.1.2), instead of properly symlinking to the `packages/airplane/` directory. The symlinking works if the version here exactly matches the version in `packages/airplane/package.json`.

.github/CONTRIBUTING.md
```diff
@@ -2,4 +2,4 @@
 
 ## Deployment
 
-To release a new version of the SDK, open a PR that bumps the version in `package.json`. Once merged to `main`, this will trigger a publish via GitHub Actions. This will publish the package to npm and create a GitHub release with auto-generated release notes. If the version is a pre-release (e.g. `0.2.0-4`), it will tag with `@next`. Otherwise, it will tag with `@latest`.
+To release a new version of the SDK, run `yarn version --no-git-tag-version` from the root, open a PR with the version bumps. Once merged to `main`, this will trigger a publish via GitHub Actions. This will publish the package to npm and create a GitHub release with auto-generated release notes. If the version is a pre-release (e.g. `0.2.0-4`), it will tag with `@next`. Otherwise, it will tag with `@latest`.
```
colinking, line 5:
```suggestion
To release a new version of the SDK, run `yarn version --no-git-tag-version` from the root and then open a PR with the version bumps. Once merged to `main`, this will trigger a publish via GitHub Actions. This will publish the package to npm and create a GitHub release with auto-generated release notes. If the version is a pre-release (e.g. `0.2.0-4`), it will tag with `@next`. Otherwise, it will tag with `@latest`.
```

.github/workflows/tests.yml
```diff
@@ -87,8 +98,10 @@ jobs:
...
-          gh release create "v${VERSION}" --generate-notes ${GH_ARGS} --target $(git rev-parse HEAD)
-          echo "::endgroup::"
-
+  release-notes:
+    runs-on: ubuntu-latest
+    concurrency: release-notes
+    needs: [publish]
+    steps:
+      - uses: actions/checkout@v3
+      - run: gh release create "v${VERSION}" --generate-notes ${GH_ARGS} --target $(git rev-parse HEAD)
```
colinking, line 107: This is missing `GH_ARGS` from above. Same with `VERSION`. Maybe add a step that computes and exports that as output via `::set-output`?

.github/CONTRIBUTING.md
colinking, line 5: Do you need the `--no-git-tag-version` if it's set by `yarn version`?

packages/workflow-runtime/package.json
```diff
@@ -0,0 +1,22 @@
...
+  },
+  "airplane": {
+    "root": "../.."
+  }
```
colinking, line 21: We shouldn't need this, unless you're using it for examples?

packages/workflow-runtime/package.json
colinking: What about `file:../airplane`?

packages/workflow-runtime/package.json
colinking: What if you place the examples in `packages/examples/...` instead?

colinking:
@calufornia what does the file structure look like for both packages? As in the tree output we were looking at the other day.

colinking:
This LGTM, though I'll defer to @leeweisberger for a review on the general setup of yarn workspaces.

colinking (approved):

calufornia:
> @calufornia what does the file structure look like for both packages? As in the tree output we were looking at the other day.

@colinking You can take a look here! https://github.com/airplanedev/node-sdk/runs/7642962425?check_suite_focus=true

.github/CONTRIBUTING.md
calufornia: Yep - the `--no-git-tag-version` flags in the `version` script only apply to the individual packages - `yarn version` applies to the root `package.json` and will attempt to create/tag a commit if we don't also include the flag at the top-level.

packages/workflow-runtime/package.json
calufornia: That somewhat works - yarn install produces a copy of the `airplane` directory, instead of a symlink:
```
ls -la node_modules/
total 0
drwxr-xr-x   3 justinlu  staff   96 Aug  2 21:46 .
drwxr-xr-x   9 justinlu  staff  288 Aug  2 21:49 ..
drwxr-xr-x  18 justinlu  staff  576 Aug  2 21:49 airplane
```
We would also need to swap out the dependency to point to the actual version of airplane before publishing, right? FWIW, it looks like Vercel also just uses static versions of to reference other vercel packages: https://github.com/vercel/vercel/commit/543ffdfe5ccfb59c9c455fc50df3bb8ed1eaf567#diff-6e2e2a1851648938b325ba84de634407a4e69a644ea61102df15ca4a8a7a9758

cc @leeweisberger if you know the best practice here

leeweisberger: Also can't figure this out. My guess is it's because of our beta package thing - this definitely works in views.

leeweisberger (changes requested):
I think we should push scripts and (most) dependencies into the children packages. That is more standard and makes the packages more self-reliant.

For example, if we were to add views to this monorepo (which I think we should) it would have different linting / testing / scripts (posssibly) because it is a react project.

packages/workflow-runtime/package.json
colinking: Looks like yarn v2 has a `workspace:*` option that should work here: https://yarnpkg.com/features/workspaces#workspace-ranges-workspace

Been wondering for awhile now if we should be moving to modern npm or yarn...

leeweisberger: Yeah we could try yarn 2. It doesn't seem to be super highly used, but yarn is basically closing bugs and mostly working on yarn 2

package.json
```diff
@@ -1,21 +1,20 @@
...
+  "private": true,
   "scripts": {
-    "test": "jest --testTimeout 500",
-    "test:watch": "jest --watch --testTimeout 500",
-    "build": "rm -rf dist && mkdir dist && cp -t dist LICENSE README.md package.json && tsc -p tsconfig.build.json",
-    "lint": "eslint --ext js,ts,yml,yaml,json --max-warnings=0 .",
-    "lint:fix": "eslint --ext js,ts,yml,yaml,json --max-warnings=0 --fix ."
+    "lint": "./node_modules/.bin/eslint --ext js,ts,yml,yaml,json --max-warnings=0 .",
+    "lint:fix": "./node_modules/.bin/eslint --ext js,ts,yml,yaml,json --max-warnings=0 --fix .",
+    "test": "./node_modules/.bin/jest --testTimeout 500",
```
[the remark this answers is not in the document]
calufornia: I was primarily following the structure of the Vercel monorepo: https://github.com/vercel/vercel/blob/e2d76e9c927f95024c5d72ea7ac492badbd7f594/package.json#L46, but per-package linting/testing configs makes sense too.

packages/airplane/package.json
```diff
@@ -0,0 +1,50 @@
...
+  "version": "0.2.0-22",
+  "description": "Node.js SDK for writing Airplane.dev tasks",
+  "main": "index.js",
+  "types": "index.d.ts",
+  "repository": "airplanedev/node-sdk",
+  "author": "hello@airplane.dev",
+  "license": "MIT",
+  "scripts": {
+    "build": "rm -rf dist && mkdir dist && cp ../../LICENSE ../../README.md package.json dist && tsc -p tsconfig.build.json",
+    "lint": "./node_modules/.bin/eslint --ext js,ts,yml,yaml,json --max-warnings=0 .",
```
leeweisberger, line 12: You probably don't need ./node_modules/.bin. Just eslint shouuuuld work. Same with jest

leeweisberger (approved):

PR #93, opened 2022-08-04
Fix release-notes job

calufornia:
The `release-notes` job needs permission to write to the Node SDK repo and also needs access to a `GITHUB_TOKEN` env var.


.github/workflows/tests.yml
```diff
@@ -103,7 +100,12 @@ jobs:
...
     concurrency: release-notes
     needs: [publish]
+    permissions:
+      # Grant the GITHUB_TOKEN permission to create releases.
+      contents: write
     steps:
       - uses: actions/checkout@v3
-      - run: |
+      - env:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
```
colinking, line 109: Do you need the token above, too, or just here?

colinking (approved):

PR #95, opened 2022-08-09
Add airplane as peer dependency

calufornia:
`airplane` is a dependency of `@airplane/workflow-runtime`, but we don't want to install it if the user's code already includes it as a dependency, since we want the `registerRuntime` function to operate on the same `runtimeMap`: https://github.com/airplanedev/node-sdk/blob/c1736dd6265989afb0f7c492fd0dfba6202a4d7f/packages/airplane/internal/runtime/index.ts#L27-L34

This moves `airplane` to be a peer dependency, which tells yarn/npm not to install it, but to warn if the dependency isn't satisfied by another package. We don't pin a specific version of `airplane` since users' code could use different version - we just enforce that it is above a certain version.


packages/workflow-runtime/package.json
```diff
@@ -38,8 +38,10 @@
   },
   "dependencies": {
     "@temporalio/worker": "1.0.1",
-    "@temporalio/workflow": "1.0.1",
-    "airplane": "0.2.0-24"
+    "@temporalio/workflow": "1.0.1"
+  },
+  "peerDependencies": {
+    "airplane": "^0.2.0-24"
```
leeweisberger, line 44: Might as well make this 25

leeweisberger (approved):
lgtm

PR #96, opened 2022-08-10
Add standard and workflow task examples

calufornia:
Adds an `examples/` directory that includes a standard and workflow task. We utilize yarn workspaces to resolve references using the local copy of `airplane` and `@airplane/workflow-runtime`, which means we no longer need to modify the import paths in `lib`. Also adds a `bumpversion` command to dissuade the direct use of `yarn version` (which must have a `--no-git-tag-version` flag supplied or else will create a commit).


examples/workflow/main.task.yaml
```diff
@@ -0,0 +1,14 @@
+slug: node_sdk_workflow
```
colinking, line 1: Do we get much by including a YAML and non-YAML example? I'm fine if we just include the config-as-code example. For both standard and workflow.

examples/workflow/main.task.ts
```diff
@@ -0,0 +1,70 @@
...
+//     airplane.input.text("Name"),
+//     airplane.input.longText("Reason"),
+//     airplane.input.sql("Query"),
+//     airplane.input.boolean("Verbose?"),
+//     airplane.input.date("Launch day"),
+//     airplane.input.datetime("Submitted At"),
+//     airplane.input.integer("Count"),
+//     airplane.input.number("Favorite number"),
+//   ]);
+// };
```
colinking, line 70: delete

examples/workflow/main.ts
```diff
@@ -0,0 +1,45 @@
...
+  await airplane.display.markdown(`Found **${users.length}** user(s).`);
+  await airplane.display.markdown(
+    JSON.stringify(users, null, 2).slice(0, 1000)
+  );
+
+  if (users.length === 0) {
+    throw new Error("Found no users");
+  } else if (users.length === 1) {
+    return users[0];
+  }
```
colinking, line 33: can you copy the additional error handling in this `lookupUser` func to the other one?

colinking (commented):
This is great, thanks! Excited we don't need those lib changes anymore.

colinking (approved):

examples/workflow/main.task.yaml
calufornia: Going to keep the YAML examples for now since that's what I tested with, there seems to be some issues (unrelated to yarn workspaces/dependencies) with deploying the code tasks 😅, maybe @ameyakhare can help get them working?

colinking: What issues did you run into @calufornia? Let's get those fixed before we launch config-as-code.

PR #97, opened 2022-08-11
Fix built-in output types

colinking:
Updates the output types of our built-ins, mostly to remove `| undefined | null`. Since we now throw an error when a run fails, it's not possible to get `undefined | null` from many of the built-ins. This makes it easier to type the response.


colinking:
* main
  * **PR #97** <a href="https://app.graphite.dev/github/pr/airplanedev/node-sdk/97" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/airplanedev/node-sdk/97?utm_source=stack-comment).

ameyakhare (approved):

PR #103, opened 2022-08-16
Add generics to `sql.query`

colinking:
This PR updates `sql.query` to support generics for typing the returned row outputs. This removes the need to cast the output and helps convey the `Q<...>` notation.

Unfortunately, TypeScript does not support variadic type parameters AFAICT. I added overloads to support up to five rows, but we can always extend this further if needed. I believe we _could_ do this with mapped tuple types (where we accept a single tuple-type generic), but I couldn't get that working within a timebox, so skipping.

Partially resolves TSK-152


colinking:
* main
  * **PR #103** <a href="https://app.graphite.dev/github/pr/airplanedev/node-sdk/103" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/airplanedev/node-sdk/103?utm_source=stack-comment).

packages/airplane/internal/builtins/sql.ts
```diff
@@ -10,24 +10,91 @@ export enum TransactionMode {
...
+  rows_affected: number;
+};
+
+type QueryExecOutputWithoutSentinel = {
+  rows_affected: number;
+};
+
+// Generates a query's output type. Checks for the QueryExecOutput type
+// which indicates whether the query returned rows.
+type QType<TRow> = TRow extends QueryExecOutput ? QueryExecOutputWithoutSentinel : TRow[];
```
leeweisberger, line 33: This is the only part I don't understand:
1. what is setting the key `__airplaneSentinelType` on rows? Is that something that we have to add?
2. if `TRow extends QueryExecOutput`, that means that there are rows or no rows? I guess from the code it looks like no rows. Maybe you can make it clear what structure we receive data in in the rows vs no rows case?

leeweisberger (approved):
Going to approve, but just let me know the answer to my questions!

packages/airplane/internal/builtins/sql.ts
leeweisberger: Also did we decide to keep the q1, q2 thing? I guess so?

colinking: Updated with doc comments. Let me know if it's still unclear.

I'm still planning to do a pass on the query output, but the consensus was that we wanted to introduce generics either way, so I decided to go ahead and add that.

PR #105, opened 2022-08-16
Add generics to other built-ins

colinking:
Updates all built-ins which return user-generated data to support generics for specifying those user-generated outputs.


colinking:
* main
  * **PR #105** <a href="https://app.graphite.dev/github/pr/airplanedev/node-sdk/105" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/airplanedev/node-sdk/105?utm_source=stack-comment).

leeweisberger (changes requested):
Talked offline, but let's default to a value that makes generics optional

packages/airplane/internal/builtins/mongodb.ts
```diff
@@ -18,14 +16,14 @@ export type UpdateOptions = {
   client?: ClientOptions;
 };
 
-export const find = async (
+export const find = async <TDocument>(
```
leeweisberger, line 19: Now find has to be a generic function right? I can't call `.find(` or I'll get an error. This should be `<TDocument=any>` so that it's optional. Lemme know I'm wrong b/c I'm not 100% sure the context in which these functions are used, but if they are called directly by users now generics will be required.

leeweisberger (changes requested):

packages/airplane/internal/builtins/mongodb.ts
colinking: Ah, TIL that `<TDocument>` defaults to `unknown` not `any`. Good catch! Will update.

leeweisberger (approved):

PR #106, opened 2022-08-16
Replace enums with string unions

colinking:
Replaces our enum usage with string unions. These are easier to use as you do not need to import the enum type, but they still offer similar type-safety guarantees and intellisense.

PR #107, opened 2022-08-17
Add `airplane.input.select`

colinking:
This adds a helper for selecting one element from a list of elements. This behavior can be achieved through `options`, but it's fairly verbose.

In the most simple case -- picking from a list of strings -- the usage is simple:

```ts
const id = await airplane.input.select("User ID", userIDs);
```

For objects, you'll likely want to add a custom label:

```ts
const user = await airplane.input.select("User", users, { optionToLabel: user => user.name });
```

Either way, this implementation avoids type issues you encounter when using `airplane.input.text`'s `options`, where TS won't know that the returned value is guaranteed to be in the original list of options.

Resolves TSK-5


colinking:
* main
  * **PR #105** <a href="https://app.graphite.dev/github/pr/airplanedev/node-sdk/105" target="_blank">Graphite</a> 
    * **PR #106** <a href="https://app.graphite.dev/github/pr/airplanedev/node-sdk/106" target="_blank">Graphite</a> 
      * **PR #107** <a href="https://app.graphite.dev/github/pr/airplanedev/node-sdk/107" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/airplanedev/node-sdk/107?utm_source=stack-comment).

kellymtrinh (approved):
Nice!!

PR #116, opened 2022-08-23
Expose Fetcher for public usage

geodn:


leeweisberger (changes requested):
What's the usecase here? Can you add a description? 

Why expose the fetcher rather than wrap endpoints (like we do for execute)?

PR #117, opened 2022-08-23
Add async execute and runbook methods

zhan-airplane:
This PR does the following:
1) Adds runbook execute and session get methods. Should be basically the same as the task methods, except with no outputs
2) Adds async versions of task and runbook execute methods. This basically starts the run/session and returns the runID/sessionID, instead of waiting for it to finish.

I'm not familiar enough with the temporal side of things, so I left the runbook execute methods blank on that side of things for now. I'll find time to go around and fix them up in a future PR - I mainly need the runbook/async stuff for views rn.

I'm in the process of testing these by just writing a bunch of node tasks that use these locally, but if tasks folks have a better way of testing things (esp unit/integration tests), lmk!

---
cc @stepzhou @leeweisberger @rose-huang @colinking


packages/workflow-runtime/internal/index.ts
```diff
@@ -97,6 +98,34 @@ export const runtime: RuntimeInterface = {
...
+    return executeTaskActivity(slug, params, resources, opts);
+  },
+
+  executeRunbook: async (
+    slug: string,
+    params: ParamValues = {},
+    opts: ClientOptions = {}
+  ): Promise<Session<typeof params>> => {
+    // TODO(zhan): implement.
+    throw new Error("unimplemented");
```
calufornia, line 117: Yeah the runbook implementation is likely similar to the task implementation above, but we'd just need send a session termination signal when the last block finishes - I don't imagine we'll need this for workflow tasks anytime soon.

calufornia (approved):
In terms of testing, I think you can probably add a couple [here](https://github.com/airplanedev/node-sdk/blob/main/packages/airplane/internal/execute.test.ts) for `executeAsync`/`executeRunbook`/`executeRunbookAsync` (e.g. make sure the async ones aren't calling `/v0/runs/get/...`)

PR #119, opened 2022-08-29
Support node 17

stepzhou:
Node 17 is past it's EOL but support still support it for a while since it's only a few months past its EOL

See: https://airplanedev.slack.com/archives/C01DYGWC3P0/p1660944792973649?thread_ts=1660942314.735709&cid=C01DYGWC3P0

Resolves VIEW-11


calufornia:
You'd probably want to modify `packages/airplane/package.json`, not `@airplane/workflow-runtime/package.json` right (although it might not hurt to put it in both)? My understanding is that the issue is with installing `@airplane/views` which installs `airplane`, not the workflow runtime.

packages/workflow-runtime/package.json
```diff
@@ -12,7 +12,7 @@
     "test": "jest --testTimeout 500 --passWithNoTests"
   },
   "engines": {
-    "node": "^14 || ^16 || ^18"
+    "node": "^14 || ^16 || ^17 || ^18"
```
calufornia, line 15: You'd probably want to modify `packages/airplane/package.json`, not `@airplane/workflow-runtime/package.json` right (although it might not hurt to put it in both)? My understanding is that the issue is with installing `@airplane/views` which installs `airplane`, not the workflow runtime.

calufornia (commented):
It looks like this probably just works without any extra modification to lib, since the view builder just [always uses the Node 18 base image](https://github.com/airplanedev/lib/blob/8a0a57f08f6b733eb7a23c30a7206351013a0428/pkg/build/view.go#L36). However, this means that if they're using Node 17, they'd only be able to develop views in the project, not tasks, since we don't support 17 for node tasks (that would require changes to lib to add the node 17 base image, modify schema, etc.)

packages/workflow-runtime/package.json
stepzhou: Woops that was my bad

stepzhou:
> that would require changes to lib to add the node 17 base image, modify schema, etc.

@calufornia What happens if users try to develop tasks with node 17? Do they just get an error?

calufornia (approved):
cc @colinking

calufornia:
> > that would require changes to lib to add the node 17 base image, modify schema, etc.
> 
> @calufornia What happens if users try to develop tasks with node 17? Do they just get an error?

I think you can still technically dev tasks with an unsupported node version (e.g. 17) through `airplane dev` - you just wouldn't be able to deploy a task with `nodeVersion: 17` since it would fail on schema validation.

PR #124, opened 2022-09-01
Configure dual ESM and CommonJS exports

colinking:
While looking into a v0.2 release of the JS SDK, I took a look enabling dual exports. I found some good documentation on dual exports here: https://antfu.me/posts/publish-esm-and-cjs

There are a few libraries that help with building TS packages. It looks like tsdx is no longer maintained, unless it gets revived as part of Turborepo. The modern alternatives appear to be [tsup](https://tsup.egoist.dev/), [microbundle](https://github.com/developit/microbundle), and [unbuild](https://github.com/unjs/unbuild).

You can test it by running `npx unbuild` in `packages/airplane` however it currently fails while generating `.d.ts` files. There is also a warning about mixing default and named exports that we'll need to dig further into. Putting on hold for now.

cc @leeweisberger 

Resolves TSK-215


colinking:
* main
  * **PR #124** <a href="https://app.graphite.dev/github/pr/airplanedev/node-sdk/124" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/airplanedev/node-sdk/124?utm_source=stack-comment).

PR #125, opened 2022-09-06
Retry execute task activity when parent run hasn't been created yet

calufornia:
Occasionally, (Airplane) workflows fail because the (Temporal) workflow starts before the Airplane run has been created. This happens in the case of task composition - if the user's code involves task composition, the child run will start executing before the parent run even exists, because the Temporal workflow is started immediately. This happens more often when a task is executed repeatedly (e.g. on a schedule every minute), and there is a Temporal worker already up and running that can start progress on the workflow.

Also modifies the `HTTPError` type and attempts to parse a custom airplane code (different from the status code) to better handle errors.

This PR makes the `executeTaskActivity` retryable so that in the case that the parent run doesn't exist yet, we invoke a backoff before trying again, to handle this scenario.


packages/workflow-runtime/internal/index.ts
```diff
@@ -45,13 +57,23 @@ export const createActivities = () => ({
...
+  startToCloseTimeout: "120s",
+  retry: {
+    maximumAttempts: 1,
+  },
+});
+
+const { executeTaskActivity } = proxyActivities<ReturnType<typeof createActivities>>({
+  startToCloseTimeout: "120s",
+  retry: {
+    initialInterval: "3s",
```
colinking, line 72: IIUC, this will wait 3s, 6s, 12s, 24s, then 48s between retries? Just confirming.

colinking (approved):
AD: let's use `code`

Re-request if you want another pair of eyes

packages/workflow-runtime/internal/index.ts
```diff
@@ -29,9 +30,9 @@ export const createActivities = () => ({
     try {
       return await client.executeTask(slug, params, resources);
     } catch (err) {
-      if (err instanceof Error) {
-        if (err.message.includes("Parent run not found")) {
-          throw ApplicationFailure.retryable(err.message, "ParentRunNotFoundError");
+      if (err instanceof HTTPError) {
+        if (err.code === "ParentRunNotFound") {
+          throw ApplicationFailure.retryable(err.message);
```
colinking, line 35: Can you add a comment here to explain why we retry this error?

packages/workflow-runtime/internal/index.ts
```diff
@@ -29,9 +30,9 @@ export const createActivities = () => ({
...
         } else {
           throw ApplicationFailure.nonRetryable(err.message, "ExecuteTaskError");
```
colinking, line 37: Why wrap these as non-retryable, but everything else is a standard `throw err`?

colinking (approved):

packages/workflow-runtime/internal/index.ts
calufornia: Following the example [here](https://docs.temporal.io/typescript/handling-failure/#pattern-wrapping-errors-with-interceptors), it seems like we could just `throw err` and it won't be retried. Wrapping this in a `nonRetryable` (when the error object has a message prop) just to be more explicit when possible.

PR #127, opened 2022-09-07
Mark sideEffects: false

colinking:
Set `sideEffects: false` to indicate that these packages are safe to tree-shake. https://webpack.js.org/guides/tree-shaking/

This will be a no-op since we do not export ESM, and AFAICT this feature only works when building ESM dependencies. Adding anyway in case I'm wrong and we're adding ESM dual exports soon anyway, so it'll eventually help. See: https://github.com/airplanedev/js-sdk/pull/124


colinking:
* colin-09-06-Directly_export_new_methods
  * **PR #127** <a href="https://app.graphite.dev/github/pr/airplanedev/node-sdk/127" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/airplanedev/node-sdk/127?utm_source=stack-comment).

PR #131, opened 2022-09-07
Rename execute*Async to execute*Background

colinking:
This PR renames `executeAsync` and `executeRunbookAsync` to `executeBackground` and `executeRunbookBackground`, respectively.

This change is meant to clarify that the "async" in the method name didn't refer to JavaScript's async/await notation. Both styles of execute methods use `async/await`, but the background variation returns immediately and let's the end-user poll for completion.

I considered `start` and `create` as alternatives, but worried that they may be too clever -- that they would be hard to distinguish from `execute`. I'm still open to naming suggestions.

I will update the Views SDK in parallel: https://github.com/airplanedev/views/pull/258

I chatted with @zhan-airplane about this change over Slack.


colinking:
* main
  * **PR #131** <a href="https://app.graphite.dev/github/pr/airplanedev/node-sdk/131" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/airplanedev/node-sdk/131?utm_source=stack-comment).

packages/airplane/internal/execute.ts
```diff
@@ -11,12 +11,12 @@ export const execute = async <Output = unknown>(
   return executeWithResources(slug, params, {}, opts);
 };
 
-export const executeAsync = async (
+export const executeBackground = async (
```
leeweisberger, line 14: Everything that we are planning on exporting externally should have a docstring imo. This gives it documentation when hovered over

leeweisberger (approved):

packages/airplane/internal/execute.ts
colinking: Agreed. We're planning to do a pass through the SDK and add docstrings to all exports, but for now I'm focused on the breaking changes that we need to land before cutting `0.2`.

PR #136, opened 2022-09-07
Add airplane.workflow 

kellymtrinh:
Instead of using `airplane.task({runtime: workflow|standard})` and specifying the runtime, users should use `airplane.task()` or `airplane.workflow()`.

Depends on https://github.com/airplanedev/lib/pull/262

CC @ameyakhare 

# Test Plan
- [x] verified that standard tasks are unaffected (I tested out `examples/standard`) and still run.

- [x] I deployed `examples/workflow/main.airplane.ts` from this branch which uses the new `airplane.workflow` locally with the lib changes,  and ran it to make sure it worked.

Screen Shot 2022-09-08 at 11 43 06 AM


packages/airplane/index.ts
```diff
@@ -51,6 +51,7 @@ export default {
   prompt,
   task,
   view,
+  workflow,
```
colinking, line 54: Pull in `main` -- we're now exporting `task/view` above too.

colinking (approved):

PR #141, opened 2022-09-09
Add default value for sql.query

colinking:
To pass the output of a SQL query into another function that expects a list of rows, you end up needing to provide a generic to `sql.query` or use a type cast. This is unfortunate, since in most cases you have a list of rows.

We avoid this by setting a more reasonable default on `sql.query`.

Before:

CleanShot 2022-09-09 at 14 47 20

After:

CleanShot 2022-09-09 at 14 48 41


colinking:
* main
  * **PR #141** <a href="https://app.graphite.dev/github/pr/airplanedev/node-sdk/141" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/airplanedev/node-sdk/141?utm_source=stack-comment).

leeweisberger (approved):

PR #159, opened 2022-10-07
Support tracing for workflows and bring workflow code into SDK

yolken-airplane:
## Description
This change adds support for tracing in workflows. This is done by registering a new otel trace provider when a workflow worker starts up, and adding traces on some of the expensive calls in the SDK client; this extra instrumentation should be a no-op if tracing hasn't been enabled at process startup (i.e., in the non-workflow case or when running a workflow with the current shims).

This change also brings in the interceptor and worker code from lib (with some cleanups and types added) so that we can make changes to these without pushing new lib versions. A follow-up change will update lib to import this functionality from the SDK and remove the excess shim code.

cc: @calufornia 

## Testing
Testing completed successfully via deploying regular and workflow tasks and verifying that traces were processed correctly in the workflow case (with the lib changes that use the new SDK functionality here).


packages/workflow-runtime/package.json
```diff
@@ -38,6 +38,12 @@
     "typescript": "4.8.4"
   },
   "dependencies": {
+    "@opentelemetry/api": "^1.2.0",
+    "@opentelemetry/exporter-trace-otlp-http": "^0.33.0",
+    "@opentelemetry/resources": "^1.7.0",
+    "@opentelemetry/sdk-node": "^0.33.0",
+    "@opentelemetry/semantic-conventions": "^1.7.0",
+    "@temporalio/interceptors-opentelemetry": "^1.4.0",
```
colinking, line 46: Pin these dependencies. Renovate will otherwise open a follow-up PR to do that.

```suggestion
    "@temporalio/interceptors-opentelemetry": "1.4.0",
```

packages/airplane/package.json
```diff
@@ -43,6 +43,7 @@
     "typescript": "4.8.4"
   },
   "dependencies": {
+    "@opentelemetry/api": "^1.2.0",
```
colinking, line 46: Pin this dependency, too.

packages/airplane/internal/api/fetcher.ts
```diff
@@ -84,7 +85,11 @@ export class Fetcher {
     );
   }
 
-  async get<Output = unknown>(path: string, params?: Record<string, unknown>): Promise<Output> {
+  async get<Output = unknown>(
+    path: string,
+    params?: Record<string, unknown>,
+    enableTracing?: boolean
```
colinking, line 91: Make this an options object. This makes it more clear what `true` means. It also allows us to introduce new options.

```suggestion
    opts: { enableTracing?: boolean } = {}
```

packages/airplane/internal/api/fetcher.ts
```diff
@@ -109,19 +114,44 @@ export class Fetcher {
...
-    const response = await this.fetch(url.toString(), {
-      method,
-      headers,
-    });
 
-    if (response.status >= 200 && response.status < 300) {
-      return (await response.json()) as Output;
-    }
+    if (enableTracing) {
+      const tracer = otel.trace.getTracer("sdk-tracer");
```
colinking, line 119: This accepts an optional version argument. Should we pass through the SDK version? You can get it from `pkg.version`. If so, should we name it `js-sdk` instead?

packages/airplane/internal/api/fetcher.ts
```diff
@@ -109,19 +114,44 @@ export class Fetcher {
...
+      return await tracer.startActiveSpan("fetcher.get", async (span) => {
```
colinking, line 120: Should we use `tracer.startSpan()` instead? By not using a closure, we wouldn't need to duplicate the `this.fetch` call.

packages/airplane/internal/api/fetcher.ts
```diff
@@ -109,19 +114,44 @@ export class Fetcher {
...
+        span.setAttribute("http_route", path);
+        span.setAttribute("http_method", method);
+        otel.propagation.inject(otel.context.active(), headers);
```
colinking, line 123: What does this do? Is it setting an attribute for each header?

packages/airplane/internal/api/client.ts
```diff
@@ -38,11 +38,15 @@ export class Client {
...
-      resources: resources ?? {},
-    });
+    }>(
+      "/v0/tasks/execute",
+      {
+        slug: slug,
+        paramValues: params ?? {},
+        resources: resources ?? {},
+      },
+      true
```
colinking, line 48: Currently, `enableTracing` defaults to `undefined`. Should it default to `true` instead? Is there anywhere where we set it to `false`?

packages/workflow-runtime/internal/interceptors.ts
```diff
@@ -0,0 +1,136 @@
...
+// See https://docs.temporal.io/docs/typescript/interceptors for details.
+class WorkflowLogOutboundInterceptor implements WorkflowOutboundCallsInterceptor {
+  info;
+  constructor(info: WorkflowInfo) {
+    this.info = info;
+  }
+
+  async scheduleActivity(
+    input: ActivityInput,
+    next: Next<WorkflowOutboundCallsInterceptor, "scheduleActivity">
```
colinking, line 33: Types -- nice!

packages/workflow-runtime/internal/interceptors.ts
```diff
@@ -0,0 +1,136 @@
...
+      workflowLog(this.info, `Signal result: ${logStringify(result)}`);
+      return result;
+    } catch (error) {
+      workflowLog(this.info, `Error handling signal: ${error}`);
+      throw error;
+    }
+  }
+}
+
+export const workflowInterceptors = () => ({
```
colinking, line 118: nit: since it's already in a package called `workflow-runtime`

```suggestion
export const interceptors = () => ({
```

packages/workflow-runtime/internal/worker.ts
```diff
@@ -0,0 +1,200 @@
...
+};
+
+// eslint-disable-next-line @typescript-eslint/no-explicit-any
+const formatWorkflowLog = (source: string, workflowInfo: WorkflowInfo, message: any) => {
+  return `airplane_workflow_log:${source}//${workflowInfo.workflowId}/${workflowInfo.runId} ${message}`;
+};
+
+// Interceptor that allows us to add extra logs around when activities start and
+// end. See https://docs.temporal.io/docs/typescript/interceptors for details.
+export class ActivityLogInboundInterceptor implements ActivityInboundCallsInterceptor {
```
colinking, line 170: Can this be unexported?

packages/workflow-runtime/internal/worker.ts
```diff
@@ -0,0 +1,200 @@
...
+        fn: (workflowInfo: WorkflowInfo, message?: any, ...optionalParams: any[]) => {
+          console.log(message, ...optionalParams);
+        },
+      },
+    },
+  };
+
+  const worker = await Worker.create({
+    workflowBundle: {
+      codePath: "/airplane/.airplane/workflow-bundle.js",
```
colinking, line 140: We shouldn't make any assumptions about paths here. Let's make this a parameter?

packages/workflow-runtime/internal/worker.ts
```diff
@@ -0,0 +1,200 @@
...
+import { WorkflowInfo } from "@temporalio/workflow";
+
+// Activity code runs in the same node process as the worker, so we import it here directly.
+import { createActivities, logStringify } from "./index";
+
+// Main worker entrypoint; starts a worker that will process activities
+// and workflows for a single task queue (equivalent to airplane task revision).
+//
+// eslint-disable-next-line @typescript-eslint/no-explicit-any
+export const runWorker = async (params: any) => {
```
colinking, line 28: Is `params` used?

packages/workflow-runtime/internal/worker.ts
```diff
@@ -0,0 +1,200 @@
...
+      ],
+    },
+    sinks,
+  });
+
+  await worker.run();
+};
+
+// eslint-disable-next-line @typescript-eslint/no-explicit-any
+const formatWorkflowLog = (source: string, workflowInfo: WorkflowInfo, message: any) => {
```
colinking, line 164: Can we combine the various logger utilities? There's `formatWorkflowLog`, `activityLog` (2x), and `workflowLog`. If not, can we at least move them to the same file? `workflow-runtime/internal/logs.ts`

packages/workflow-runtime/internal/activities.ts
```diff
@@ -134,3 +135,12 @@ const deserializeTemporalError = (err: unknown) => {
 
   return err;
 };
+
+// eslint-disable-next-line @typescript-eslint/no-explicit-any
+export const activityLog = (message?: any, ...optionalParams: any[]) => {
```
colinking, line 140: Is this used? There's another `activityLog` below.

packages/workflow-runtime/internal/index.ts
```diff
@@ -155,3 +157,12 @@ const passthroughOptions = (opts: ClientOptions): ClientOptions => {
 export const registerWorkflowRuntime = () => {
   registerRuntime("workflow", runtime);
 };
+
+const sensitiveLogKeys = ["airplane_token", "airplane_resources", "password", "token"];
+
+// eslint-disable-next-line @typescript-eslint/no-explicit-any
+export const logStringify = (obj: any): string => {
```
colinking, line 164: nit since it doesn't do any logging. Does this need to be exported from the `index` file? If not, let's put it in `worker.ts` and then the interceptors file can import it. Let's try to keep unnecessary exports out of the `index.ts` file.

```suggestion
export const stringifyOmitSensitive = (obj: any): string => {
```

colinking (changes requested):
Looks good. Fairly minor feedback.

packages/workflow-runtime/package.json
yolken-airplane: Fixed.

packages/airplane/internal/api/fetcher.ts
yolken-airplane: Done.

packages/airplane/internal/api/fetcher.ts
yolken-airplane: Fixed.

packages/airplane/internal/api/fetcher.ts
yolken-airplane: It adds a bunch of standardized headers. Added a comment with a link to the spec.

packages/airplane/internal/api/fetcher.ts
yolken-airplane: So the benefit of `startActiveSpan` is that the tracing context is nested, which means that if we instrument `fetch` in the future, it will inherit the parent from the outer span. This is the approach recommended by the otel docs unless you need to create multiple parallel spans within the same function.

packages/airplane/internal/api/client.ts
yolken-airplane: I'm leaving it undefined (false) for the various GET requests we make including `/v0/runs/get`, `/v0/sessions/get`, etc. I'd prefer to leave it as defaulting to false/undefined so we don't accidentally overwhelm Honeycomb if we add a new, frequently accessed API endpoint. :-)

packages/workflow-runtime/internal/interceptors.ts
yolken-airplane: Yeah, they help a lot. :-)

packages/workflow-runtime/internal/interceptors.ts
yolken-airplane: Fixed.

packages/workflow-runtime/internal/worker.ts
yolken-airplane: Yup, done.

packages/workflow-runtime/internal/worker.ts
yolken-airplane: Nope. Removed it and replaced with a single argument for the code path.

packages/workflow-runtime/internal/worker.ts
yolken-airplane: Yeah, good idea. I've moved everything into a single file and also cleaned up the naming a bit so it's more clear what is used where.

packages/workflow-runtime/internal/activities.ts
yolken-airplane: Moved to the logging file and renamed the other one to make the usage more clear.

packages/workflow-runtime/internal/index.ts
yolken-airplane: Fixed.

packages/airplane/internal/api/fetcher.ts
colinking: Which direction is this injecting -- is it updating `headers` with additional tracing headers, or is it adding `headers` onto the trace context? I interpreted as the latter originally. Assuming the former, a small tweak:

```
// Attaches trace context headers onto `headers`
// See https://www.w3.org/TR/trace-context for more details.
```

packages/workflow-runtime/internal/logging.ts
```diff
@@ -0,0 +1,47 @@
+import { Context, Info } from "@temporalio/activity";
```
colinking, line 1: nice :)

colinking (approved):
looks great. thanks for incorporating all that!

colinking:
@yolken-airplane when you merge, there are deployment instructions here: https://github.com/airplanedev/js-sdk/blob/main/.github/CONTRIBUTING.md

yolken-airplane:
> Which direction is this injecting -- is it updating headers with additional tracing headers, or is it adding headers onto the trace context? I interpreted as the latter originally. Assuming the former, a small tweak:

Yeah, it's the former. Updated the comment based on your suggestion.

packages/airplane/internal/api/fetcher.ts
yolken-airplane: Fixed.

PR #161, opened 2022-10-12
Bump version to 0.3.0

yolken-airplane:
## Description
This change bumps the version of the js-sdk to 0.3.0 to pick up the changes from https://github.com/airplanedev/js-sdk/pull/159. The changes here are bigger than bug fixes, so figured it made sense to reset the minor version.


colinking (changes requested):
For `0.x` versions, the minor version is used to indicate breaking changes. Since your PR didn't make any breaking changes (CMIIW?), we should stick to bumping the patch version.

yolken-airplane:
> For `0.x` versions, the minor version is used to indicate breaking changes. Since your PR didn't make any breaking changes (CMIIW?), we should stick to bumping the patch version.

It's semi-breaking since once we update lib, the old versions won't work any more?

PR #164, opened 2022-10-13
Add AirplaneFile abstraction

geodn:
Babel is needed to get fetch-blob to work, and fetch-blob is needed as a browser- and node-compatible implementation of the Blob API.

AirplaneFile is a class that can be used in both views and nodejs tasks/workflows. The plan is to provide convenience factory functions that can construct AirplaneFile from url, from blob, etc.


packages/airplane/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "airplane",
-  "version": "0.2.10",
+  "version": "0.2.11",
```
colinking, line 3: You have to run `yarn bumpversion` to properly bump the version. That bumps the version in the rest of the workspace.

packages/airplane/index.ts
```diff
@@ -37,6 +39,7 @@ export {
 // airplane.execute(...)
 // airplane.email.message(...)
 export default {
+  AirplaneFile,
```
colinking, line 42: Let's not include it in the default export. Since the name already includes `Airplane...`, we're already prefixing it such that it can be imported directly.

IOW, we should both default and directly export all methods, but only directly export types and classes. wdyt?

packages/airplane/internal/file/file.ts
```diff
@@ -0,0 +1,14 @@
+import { Blob } from "fetch-blob";
+
+export class AirplaneFile extends Blob {
+  public id: string;
+  public url: string;
+  public name: string;
+
+  constructor(id: string, url: string, name: string, source: Blob) {
```
colinking, line 8: Having multiple parameters in a row that have the same type makes the callsites hard to read. Let's use an object instead.

```ts
constructor(source: Blob, metadata: { id: string, url: string, name: string })
```

We also landed on not including the `name` field, right?

packages/airplane/internal/file/file.ts
```diff
@@ -0,0 +1,14 @@
+import { Blob } from "fetch-blob";
+
+export class AirplaneFile extends Blob {
```
colinking, line 3: Add a docstring.

packages/airplane/babel.config.js
```diff
@@ -0,0 +1,12 @@
+module.exports = {
+  presets: [
+    [
+      "@babel/preset-env",
```
colinking, line 4: Add comments explaining why we need this.

packages/airplane/internal/file/file.test.ts
```diff
@@ -0,0 +1,11 @@
+import { Blob } from "fetch-blob";
```
colinking, line 1: Add a comment describing why we use this `Blob` instead of the native Node.js `Blob`.

packages/airplane/babel.config.js
```diff
@@ -0,0 +1,12 @@
+module.exports = {
```
colinking, line 1: We also have a `babel.config.js` in the top-level. Should we place this there? https://github.com/airplanedev/js-sdk/blob/main/babel.config.js

colinking, line 1: How is this babel config getting used? We configure Jest to use babel, but don't configure babel anywhere else AFAIK.

colinking (changes requested):

packages/airplane/internal/file/file.test.ts
```diff
@@ -0,0 +1,11 @@
...
+
+import { AirplaneFile } from "./file";
+
+describe("AirplaneFile", () => {
+  test("works in node", async () => {
```
colinking, line 6: Looks like you might be able to add a jsdom test, too, but using this: https://jestjs.io/docs/configuration#testenvironment-string

packages/airplane/internal/file/file.test.ts
colinking, line 1: Good find on `fetch-blob` :)

packages/airplane/package.json
geodn: Done.

packages/airplane/index.ts
geodn: Done.

packages/airplane/internal/file/file.ts
geodn: `name` ends up needed in order to show a card for the uploaded file. My thinking was to not rely on guaranteeing that the last part of url is always the name; seems easier to me if we just set the name based on url if we don't have the name available?

packages/airplane/internal/file/file.test.ts
geodn: Added in file.ts.

packages/airplane/babel.config.js
geodn: It doesn't work when it's added there - seems like it's using the babel config from the `airplane` directory.

geodn: It's required to get tests to pass. Honestly I'm not fully sure what's happening, but I think fetch-blob is implemented with a different version of js (ES6?) and there's some compatibility issues that are resolved by babel. Here's the full error when the babel config isn't present:

```
Test suite failed to run
                                                                                                                          
    Jest encountered an unexpected token             
                                                             
    Jest failed to parse a file. This happens e.g. when your code or its dependencies use non-standard JavaScript syntax, or when Jest is not configured to support such syntax.                                                                     
                                                             
    Out of the box Jest supports Babel, which will be used to transform your files into valid JS based on your Babel configuration.                                                                                                                  
                                                             
    By default "node_modules" folder is ignored by transformers.                                                          
                                                             
    Here's what you can do:                                                                                               
     • If you are trying to use ECMAScript Modules, see https://jestjs.io/docs/ecmascript-modules for how to enable it.   
     • If you are trying to use TypeScript, see https://jestjs.io/docs/getting-started#using-typescript                   
     • To have some of your "node_modules" files transformed, you can specify a custom "transformIgnorePatterns" in your config.                                                                                                                     
     • If you need a custom transformation specify a "transform" option in your config.                                   
     • If you simply want to mock your non-JS modules (e.g. binary assets) you can stub them out with the "moduleNameMapper" config option.                                                                                                          
                                                             
    You'll find more details and examples of these config options in the docs:
    https://jestjs.io/docs/configuration            
    For information about custom transformations, see:
    https://jestjs.io/docs/code-transformation                                                                            
                                                             
    Details:                                              
                                                             
    /Users/georgedu/go/src/github.com/airplanedev/js-sdk/node_modules/fetch-blob/index.js:4
    import './streams.cjs'; // 64 KiB (same size chrome slice theirs blob into Uint8array's)                              
    ^^^^^^                                         
                                                             
    SyntaxError: Cannot use import statement outside a module 
```

packages/airplane/babel.config.js
geodn: Done.

packages/airplane/internal/file/file.test.ts
geodn: Very cool! Would we create something like file.jsdom.test.ts?

It does look like it requires a newer version of jest than the one we're using presently (29.1)

packages/airplane/babel.config.js
colinking: Good catch. AFAICT, that top-level babel config was not getting used. I pushed up a commit to remove it.

colinking (approved):
Looks great!

PR #165, opened 2022-10-17
Allow AirplaneFile to be passed to execute functions, and bump version to 0.2.12

geodn:
Tasks expect the upload id to be passed for any file fields. E.g. if a task has a parameter of type File with slug `my_file`, the params should be `{ my_file: "upload_id_of_file" }`

This change allows AirplaneFile to be passed for convenience. We rewrite the params right above the runtime level because runtimes may have any implementation, and this is the lowest possible level to make the change such that all code dependent on js-sdk for task/runbook execution will catch the change.

Testing strategy: yarn test, as well as e2e test, creating AirplaneFile with FileInput and calling useTaskQuery.


packages/airplane/internal/execute.ts
```diff
@@ -1,8 +1,19 @@
 import { ClientOptions } from "./api/client";
 import { ParamValues, Run, Session, TaskRun } from "./api/types";
 import { RunTerminationError, SessionTerminationError } from "./errors";
+import { AirplaneFile } from "./file";
 import { getRuntime } from "./runtime";
 
+const processParams = (params: ParamValues): ParamValues => {
```
colinking, line 7:
```suggestion
const normalizeParams = (params: ParamValues): ParamValues => {
```

packages/airplane/internal/api/types.ts
```diff
@@ -97,6 +99,7 @@ export type ParamType = keyof ParamJSTypes;
 export type ParamValue =
   | undefined
   | null
+  | AirplaneFile
```
colinking, line 102: You'll need to update `JSParamValues`. That's used to generate parameter value typing for inline config. You can test by creating an inline configured task and trying to call it. e.g.

```ts
import airplane from "airplane";

export const myWorkflow = airplane.workflow({
  slug: "js_sdk_workflow",
  name: "[JS SDK] Workflow",
  parameters: {
    name: {
      name: "Name",
      slug: "name",
      type: "shorttext",
      default: "Colin",
      required: false,
    },
    myFile: "file"
  },
  resources: ["db"],
}, async params => {
  console.log(`Looking up users like '${params.name}'`);
  const run = await airplane.sql.query<{id: string; name: string}>(
    "db",
    `
      select id, name from users
      where (name ilike :name or email ilike :name) and not is_suspended and not is_service_user
      order by name asc
    `,
    { args: { name: `%${params.name}%` } }
  );
  const users = run.output.Q1;

  await airplane.display.markdown(`Found **${users.length}** user(s).`);
  await airplane.display.json(users);
  await airplane.display.table(users);

  if (users.length === 0) {
    throw new Error("Found no users");
  } else if (users.length === 1) {
    return users[0];
  }

  myWorkflow({
    // This should accept an AirplaneFile:
    myFile: "tes"
  })

  const user = await airplane.prompt.select("User", users, {
    default: users[0],
    optionToLabel: u => u.name,
  })

  return user
})
```

colinking (approved):

PR #171, opened 2022-10-25
Update JS inline code config to return to a run

leeweisberger:
Update JS inline code config to return a run insted of run output so it is consistent with `airplane.execute`.

This is a breaking change, but it is in a beta product.

TSK-927


ameyakhare (changes requested):
I think you need to not throw an error if the status is not succeeded in the execute function, or do you want to maintain that behavior?

leeweisberger:
> I think you need to not throw an error if the status is not succeeded in the execute function, or do you want to maintain that behavior?

Can you explain? What does this mean and how is it related? is this a separate issue?

ameyakhare (approved):

ameyakhare:
> > I think you need to not throw an error if the status is not succeeded in the execute function, or do you want to maintain that behavior?
> 
> Can you explain? What does this mean and how is it related? is this a separate issue?
> 
> Don't both inline and non-inline call the same execute function?

this is fine, its just that a RunTerminationError is thrown in the execute call if the result status is not successful. it will return a run object only if the run succeeds, but no run object if the run fails. but you're right its a separate issue so unrelated to this

PR #172, opened 2022-10-26
Add request timeouts and check for temporal cancellationSignal

ericliuche:
This change adds support for request timeouts to the api client and also checks temporal's context cancellationSignal to allow us to early abort a request if a workflow has been cancelled. 

TSK-931

Tests
------

Ran locally and tested error, added a fake delay of 20s in the server. We correctly see the time out error and the retries in temporal.

Screen Shot 2022-11-01 at 12 19 39 PM

Without the delay, the example workflow still works.


ericliuche:
Still testing this change, but putting it up in case people have thoughts

packages/airplane/internal/api/fetcher.ts
```diff
@@ -113,76 +146,42 @@ export class Fetcher {
...
       const response = await this.fetch(url.toString(), {
         method,
         headers,
+        body: opts.body ? JSON.stringify(opts.body) : undefined,
+        signal: controller.signal,
       });
       if (response.status >= 200 && response.status < 300) {
         return (await response.json()) as Output;
       }
       throw await HTTPError.newFromResponse(method, response);
```
ericliuche, line 179: should we retry the request if it time's out? the issue here is we can't necessarily tell if the request abort was due to timeout or workflow request cancellation :/

packages/airplane/internal/api/fetcher.ts
```diff
@@ -1,4 +1,5 @@
 import otel from "@opentelemetry/api";
+import { AbortSignal } from "abort-controller";
```
colinking, line 2: Add `abort-controller` as a dependency.

packages/airplane/internal/api/fetcher.ts
```diff
@@ -88,15 +89,47 @@ export class Fetcher {
   async get<Output = unknown>(
     path: string,
     params?: Record<string, unknown>,
-    opts: { enableTracing?: boolean } = {}
+    opts: { enableTracing?: boolean; signal?: AbortSignal; timeoutMs?: number } = {}
```
colinking, line 92: nit

```suggestion
    opts: { enableTracing?: boolean; abortSignal?: AbortSignal; timeoutMs?: number } = {}
```

colinking: Maybe define this struct above and re-use it below? That way you can add doc comments.

```typescript
// ...
type RequestOptions = {
  // ...
  enableTracing?: boolean
  // ...
  abortSignal?: AbortSignal
  // ...
  timeoutMs?: number
}
```

packages/airplane/internal/api/fetcher.ts
```diff
@@ -88,15 +89,47 @@ export class Fetcher {
...
+    body?: Record<string, unknown>,
+    opts: { enableTracing?: boolean; signal?: AbortSignal; timeoutMs?: number } = {}
+  ): Promise<Output> {
+    return this.request<Output>(path, "POST", {
+      ...opts,
+      body,
+    });
+  }
+
+  async request<Output = unknown>(
```
colinking, line 111: Nice refactor.

```suggestion
  protected async request<Output = unknown>(
```

packages/airplane/internal/api/fetcher.ts
```diff
@@ -88,15 +89,47 @@ export class Fetcher {
...
+    path: string,
+    method: "GET" | "POST",
+    opts: {
+      enableTracing?: boolean;
+      params?: Record<string, unknown>;
```
colinking, line 116:
```suggestion
      queryParams?: Record<string, unknown>;
```

packages/airplane/internal/api/fetcher.ts
colinking: Could you just check `opts.signal.aborted` to see if it was a cancellation?

packages/airplane/internal/api/fetcher.ts
```diff
@@ -113,76 +146,42 @@ export class Fetcher {
...
-      headers["X-Airplane-Env-Slug"] = this.envSlug;
-    }
-    if (this.source) {
-      headers["X-Airplane-Client-Source"] = this.source;
-    }
 
-    const method = "POST";
-
-    if (opts.enableTracing) {
+    if (!opts.enableTracing) {
```
colinking, line 182: 😄 

```suggestion
    if (opts.enableTracing) {
```

packages/airplane/internal/api/fetcher.ts
```diff
@@ -113,76 +146,42 @@ export class Fetcher {
...
       const tracer = otel.trace.getTracer("js-sdk", pkg.version);
-      return await tracer.startActiveSpan("fetcher.post", async (span) => {
+      return await tracer.startActiveSpan(`fetcher.${method.toLowerCase()}`, async (span) => {
```
colinking, line 184: Is it more idiomatic to name the span `fetch` and move the method to an attribute? If not sure, backwards-compat is fine with me.

packages/airplane/internal/api/fetcher.ts
```diff
@@ -191,30 +190,13 @@ export class Fetcher {
...
-          const response = await this.fetch(url.toString(), {
-            method,
-            body: body ? JSON.stringify(body) : undefined,
-            headers,
-          });
-          if (response.status >= 200 && response.status < 300) {
-            return (await response.json()) as Output;
-          }
-          throw await HTTPError.newFromResponse(method, response);
+          return makeRequest();
```
colinking, line 193: Otherwise you leave the `try` clause too early.

```suggestion
          return await makeRequest();
```

packages/airplane/internal/api/fetcher.ts
```diff
@@ -113,76 +146,42 @@ export class Fetcher {
...
-        otel.propagation.inject(otel.context.active(), headers);
+    if (opts.signal) {
+      if (opts.signal.aborted) {
+        onAbort();
+      } else {
+        opts.signal.addEventListener("abort", onAbort);
+      }
+    }
+    if (opts.timeoutMs) {
+      setTimeout(() => controller.abort(), opts.timeoutMs);
```
colinking, line 166: Make sure to `clearTimeout`, e.g. if the request finishes before the timeout.

packages/airplane/internal/api/fetcher.ts
```diff
@@ -113,76 +146,42 @@ export class Fetcher {
...
 
-    if (opts.enableTracing) {
-      const tracer = otel.trace.getTracer("js-sdk", pkg.version);
-      return await tracer.startActiveSpan("fetcher.get", async (span) => {
-        span.setAttribute("http_route", path);
-        span.setAttribute("http_method", method);
+    const onAbort = () => {
+      controller.abort();
+      if (opts.signal) {
+        opts.signal.removeEventListener("abort", onAbort);
```
colinking, line 154: Make sure to `removeEventListener` once the request finishes, if not aborted.

packages/airplane/internal/api/fetcher.ts
colinking, line 179: What does the error look like? We might want to catch the error here and convert it into a pretty error message.

packages/airplane/internal/api/fetcher.ts
colinking, line 166: What is the default timeout value for `fetch`? Maybe we should we always set one?

packages/airplane/internal/api/client.ts
```diff
@@ -47,6 +50,8 @@ export class Client {
       },
       {
         enableTracing: true,
+        signal: opts?.signal,
+        timeoutMs: 30 * 1000,
```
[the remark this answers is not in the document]
colinking: I don't expect the response to take much time.

packages/airplane/internal/api/client.ts
```diff
@@ -67,64 +73,74 @@ export class Client {
...
-    const { output } = await this.fetcher.get<{ output: O }>("/v0/runs/getOutputs", {
-      id: runID,
-    });
+  async getRunOutput<O = unknown>(runID: string, opts?: { signal: AbortSignal }): Promise<O> {
+    const { output } = await this.fetcher.get<{ output: O }>(
+      "/v0/runs/getOutputs",
+      {
+        id: runID,
+      },
+      { signal: opts?.signal }
```
colinking, line 90: add a timeout

packages/airplane/internal/api/client.ts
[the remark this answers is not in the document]
colinking: How about a 10s timeout by default, and bump the execute runbook timeout to 30s? (that endpoint is known to be slow).

colinking (changes requested):
Great catch on this!

packages/airplane/internal/api/fetcher.ts
colinking, line 166: Can you add a test case that covers a timeout and another that covers an abort signal? 🙏🏻

packages/airplane/internal/api/fetcher.ts
ericliuche: gonna keep as is for backwards compat for now

packages/airplane/internal/api/fetcher.ts
ericliuche: default is no timeout, set our default to 10s for now

packages/airplane/internal/api/fetcher.ts
ericliuche: The error looks like `The user aborted a request.`. Temporal seems to check `https://github.dev/temporalio/sdk-typescript/blob/eb765683e5376fa3c79d035c5e79b053ad02d706/packages/worker/src/activity.ts#L94` for the AbortError and turns it into a CancelledFailure

ericliuche: Event if we don't differentiate the timeout from the cancellation, i think it's fine. In the case that the timeout happened, we'll retry again. In the case that there's been a cancellation, the signal will stay aborted and the request  wont start.

Screen Shot 2022-10-31 at 10 33 17 AM

ericliuche:
Ah heck `AbortController` seems to have been introduced in node 15 :( is there a good way around that @colinking ?

colinking:
@ericliuche you can use a polyfill, e.g. https://www.npmjs.com/package/abort-controller

packages/airplane/internal/api/fetcher.ts
```diff
@@ -113,76 +157,50 @@ export class Fetcher {
...
-            return (await response.json()) as Output;
-          }
-          throw await HTTPError.newFromResponse(method, response);
-        } finally {
-          span.end();
+    const makeRequest = async () => {
+      const teardownSignals = () => {
+        clearTimeout(timeoutID);
+        if (opts.abortSignal) {
+          opts.abortSignal.removeEventListener("abort", onAbort);
```
colinking, line 183: fwiw you can do `opts.abortSignal?.removeEventListener(...)` instead of having the `if` check

colinking (approved):
LGTM, only remaining nit is to improve the error message before merging!

packages/airplane/internal/api/fetcher.ts
```diff
@@ -58,19 +68,20 @@ export class Fetcher {
...
+            if (error.name === "FetchError" && (error as FetchError).type === "system") {
+              // Don't retry nock matching errors (indicates an invalid tests)
+              if ((error as FetchError).errno === "ERR_NOCK_NO_MATCH") {
+                return false;
+              }
+              return true;
+            }
+            // Retry request timeouts
+            else if (error.name === "AbortError") {
+              return true;
```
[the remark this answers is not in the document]
ericliuche: in the case that a request times out, i think we do still want to retry? like if our API is just being flaky. the issue is we can't differentiate here between `request timed out due to flakiness` and `request aborted due to signal` but i think its fine even to retry the latter since calling `fetch` with an already aborted signal just immediately returns with the Abort error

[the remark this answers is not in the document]
colinking: But won't the controller be aborted in the timeout case, too?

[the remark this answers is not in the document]
ericliuche: yeah, but i thought we would want to retry timeouts?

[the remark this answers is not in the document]
colinking: These retries are happening within the `await fetch()` call. The timeout added in this PR applies across _all retries_, not an individual instance of a retry.

I'm open to retrying on timeout, but I'm just pointing out that it doesn't work as-is.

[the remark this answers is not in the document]
ericliuche: ohhhhhh

packages/airplane/internal/api/fetcher.ts
```diff
@@ -113,76 +157,50 @@ export class Fetcher {
...
+        opts.abortSignal.removeEventListener("abort", onAbort);
+      }
+    };
 
-        // Attach trace context headers onto `headers`. See
-        // https://www.w3.org/TR/trace-context/ for more details.
-        otel.propagation.inject(otel.context.active(), headers);
+    if (opts.abortSignal) {
+      if (opts.abortSignal.aborted) {
+        onAbort();
```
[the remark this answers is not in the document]
ericliuche: yeah, it sets the `controller` to aborted so the `fetch` will raise the abort error that the temporal sdk expects. Calling `removeEventListener` without a valid identifier is fine, https://developer.mozilla.org/en-US/docs/Web/API/EventTarget/removeEventListener

```
Calling removeEventListener() with arguments that do not identify any currently registered [event listener](https://developer.mozilla.org/en-US/docs/Web/API/EventTarget/addEventListener#the_event_listener_callback) on the EventTarget has no effect.
```

PR #176, opened 2022-11-01
Allow Date objects to be passed as task params

geodn:
We include the timezone information when converting Date to string, so:
1. The backend can get the exact timestamp for datetime values, and
2. The backend can get the local date for date values, avoiding off-by-one errors.

The first commit makes the change, and the second commit bumps the version. Reference: https://github.com/airplanedev/js-sdk/commit/51aa4808cc5afc1bda4df3c5ea04fa742cdfeb63


packages/airplane/internal/execute.ts
```diff
@@ -4,11 +4,27 @@ import { RunTerminationError, SessionTerminationError } from "./errors";
...
+  return `${date}T${time}${timezone}`;
+};
+
 const normalizeParams = (params: ParamValues): ParamValues => {
   const newParamEntries = Object.entries(params).map(([key, val]) => {
     if (val instanceof AirplaneFile) {
       return [key, val.id];
     }
+    if (val instanceof Date) {
+      return [key, `${toISOString(val)}`];
```
zhan-airplane, line 26: Doesn't `Date` already have `toISOString()`...? https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Date/toISOString

zhan-airplane (approved):

packages/airplane/internal/execute.ts
geodn: The built in `toISOString` doesn't include the right timezone, it always formats to UTC time.

zhan-airplane: Can you add a comment for why we need it then

packages/airplane/package.json
```diff
@@ -26,6 +26,7 @@
     "@types/node": "18.11.7",
     "@types/node-fetch": "2.6.2",
     "@types/react": "18.0.24",
+    "@types/timezoned-date": "^3.0.0",
```
colinking, line 29: FYI @geodn you have to pin these or else Renovate will open a PR to pin them for you 😄

PR #177, opened 2022-11-01
Bump version to 0.2.14

colinking:


colinking:
https://github.com/airplanedev/js-sdk/pull/176

PR #178, opened 2022-11-01
Fix open handles for fetcher test

ericliuche:
Call `nock.abortPendingRequests()` to stop the ongoing delayed nock request


colinking (approved):
nice!

PR #179, opened 2022-11-01
Escape newlines in workflow log messages

yolken-airplane:
## Description
This change escapes newlines in workflow log messages so that they're routed correctly by the runbroker. A follow-up change will unescape them downstream so they're more readable in the UI.

Resolves TSK-792

## Testing
Testing completed successfully via deploying a workflow with a multi-line log message and verifying that the entire message was visible in the UI:

```
console.log("This is a really long log line\nthat's split across\nmultiple lines.");
```

Screen Shot 2022-11-01 at 5 05 35 PM


packages/workflow-runtime/internal/logging.ts
```diff
@@ -34,7 +38,7 @@ export const workflowInterceptorLog = (info: WorkflowInfo, message: any) => {
   // Note that we cannot directly call console.log here because we override the console.log method in the
   // workflow shim, which would cause the output to get unnecessarily prepended with
   // airplane_workflow_log:workflow.
-  logger.raw(workflowLogMessage("interceptor", info, message));
+  logger.raw(workflowLogMessage("interceptor", info, message.replace(/\n/g, "\\n")));
```
colinking, line 41: this already calls `workflowLogMessage` which escapes

colinking (approved):

packages/workflow-runtime/internal/logging.ts
yolken-airplane: Oh, yeah, you're right. Fixed!

PR #194, opened 2022-11-10
Output of async task should not be a promise

leeweisberger:
I **think** I broke our typing here. The output of an async task should be a synchronous run with a synchronous output.

Please check me on this that this is correct now!


leeweisberger:
cc @yolken-airplane

packages/airplane/internal/config.test-d.ts
```diff
@@ -20,6 +20,21 @@ const _task = async () => {
     //       }
   }
 
+  // Async task with no parameters.
```
leeweisberger, line 23: This test failed without my change but now passes

ameyakhare (approved):

PR #195, opened 2022-11-10
Add constraint to examples

colinking:
This is necessary for the examples to work when deployed into our Airplane team.


examples/standard/main.airplane.ts
```diff
@@ -13,6 +13,7 @@ export default airplane.task({
     }
   },
   resources: ["db"],
```
yolken-airplane, line 15: Should we add a comment that users adopting our examples will probably want to remove it?

yolken-airplane (approved):

examples/standard/main.airplane.ts
colinking: This gets so much nicer with resource constraints 😬 Gonna actually bail on this PR, it's a pain for local testing.

PR #196, opened 2022-11-10
Output of async task should not be a promise

leeweisberger:
I think I broke our typing here. The output of an async task should be a synchronous run with a synchronous output.

Please check me on this that this is correct now!

---

(reopened b/c I borked my git history)


ameyakhare (approved):
how many times do we have to teach this lesson to you old man

PR #198, opened 2022-11-11
Refactor parameter type system

colinking:
This PR refactors the param type system to fix a few issues. Specifically, it fixes:
- If a task has no parameters configured, it now accepts both `() => {}` and `(params) => {}` as run functions. In the latter case, `params` will be typed as `Record<string, never> | undefined`. This allows you to pass in an empty object to tasks that do not accept parameters.
- Adds a type helper, `Resolve<T>` that forces the TS compiler to resolve the type `T`. This significantly improves the types provided by intellisense when hovering over a task's parameters. For example, it would convert the following:

```ts
Optional<{
  foo: string;
}, never>

// -> 

{
  foo: string;
}
```

- Optional parameters with defaults are now correctly typed as required fields when passed into a run function.

For reviewers, I'd recommend focusing on twoslash tests. There's a brief `README` in the `params` folder. I'll likely rewrite how the `Params` types work in a follow-up PR where I fix some issues with parameter types.

Partially resolves TSK-634 -- specifically, it fixes issues (1), (2), and (4) as explained above. I'll address the rest in a follow-up PR.

This will require some changes to the Views package. I've opened a PR here: https://github.com/airplanedev/views/pull/407

cc @ericliuche


colinking:
* main
  * **PR #198** <a href="https://app.graphite.dev/github/pr/airplanedev/node-sdk/198" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/airplanedev/node-sdk/198?utm_source=stack-comment).

packages/airplane/internal/config.ts
```diff
@@ -61,40 +35,47 @@ export type TaskConfig<TParams extends Params> = {
...
-    // work for some inexplicable reason.
-    (
-      params?: TParamValues[] | Record<string, never>,
+export type RunFunc<
+  TParamValues extends SerializedParamValues = Record<string, never>,
+  // eslint-disable-next-line @typescript-eslint/no-explicit-any
+  TOutput = any
+> = (
+  ...paramValues: TParamValues extends Record<string, never>
+    ? [Resolve<TParamValues>?]
```
leeweisberger, line 44: This i've never seen but I guess it means optional?

packages/airplane/internal/type_helpers.ts
```diff
@@ -0,0 +1,19 @@
+/**
+ * Resolve forces TypeScript to resolve generics into an underlying object.
+ * It helps produce more intelligible types: https://effectivetypescript.com/2022/02/25/gentips-4-display/
+ */
+// eslint-disable-next-line @typescript-eslint/ban-types
+export type Resolve<T> = T extends Function ? T : { [K in keyof T]: T[K] };
```
leeweisberger, line 6: We should use this in views.

packages/airplane/internal/params/params.ts
```diff
@@ -0,0 +1,117 @@
...
+  | "sql"
+  | "boolean"
+  | "file"
+  | "integer"
+  | "float"
+  | "date"
+  | "datetime"
+  | "configvar";
+
+export type Param<TParamValueByType extends ParamValueByType> = { type: ParamType } & (
```
leeweisberger, line 15: Why do we need this line if each of the base param types has their own type?

packages/airplane/internal/params/params.ts
```diff
@@ -0,0 +1,117 @@
...
+export type ParamOptions<TParamValue> = {
+  options?: (TParamValue | SelectOption<TParamValue>)[];
+};
+
+export type SelectOption<TParamValue> = {
+  label: string;
+  value: TParamValue;
+};
+
+export type ParamBase<TParamValue> = {
```
leeweisberger, line 91: Do we need to export this? Maybe do a quick pass and don't export private types that aren't used elsewhere so importing from these files is easier

packages/airplane/internal/params/params_input.ts
```diff
@@ -0,0 +1,71 @@
...
+            TParams[TSlug]
+          >["type"]];
+        },
+        OptionalParamSlugs<NormalizeAsObjects<TParams>>
+      >
+    >;
+
+export type InputParam = ParamType | InputParamObject;
+export type InputParams = Record<string, InputParam>;
+type InputParamObject = Param<InputParamValueByType>;
```
leeweisberger, line 45: I'm getting lost here :( can you document some of these types that are all slight variations off each other?

packages/airplane/internal/params/params_serialized.ts
```diff
@@ -0,0 +1,69 @@
...
+  type: SerializedParamType;
+  component?: SerializedParamComponent;
+  default?: SerializedParamValue;
+  constraints: {
+    optional: boolean;
+    regex?: string;
+    options?: Array<SerializedParamValue | { label: string; value: SerializedParamValue }>;
+  };
+};
+export type SerializedParams = SerializedParam[];
```
leeweisberger, line 50: I guess you're going for consistency here? Not 100% sure this is clearer vs just adding more options that the user has to choose between. But I'll leave it up to you.

packages/airplane/internal/builtins/email.ts
```diff
@@ -25,11 +26,16 @@ export const message = async (
...
   const { subject, message, client } = opts;
   return executeWithResources(
     "airplane:email_message",
-    { sender, recipients, subject: subject ?? "", message: dedent(message ?? "") },
+    {
+      sender,
+      recipients,
+      subject: subject ?? "",
+      message: dedent(message ?? ""),
+    } as unknown as InputParamValues,
```
leeweisberger, line 38: Don't have to do this in this pr, but would probably be ideal if we typed these

packages/airplane/internal/config.test-d.ts
```diff
@@ -1,86 +1,287 @@
...
+    type _response = Awaited<ReturnType<typeof myTask>>;
+    //    ^? type _response = {
+    //           id: string;
+    //           taskID?: string | undefined;
+    //           paramValues: {
+    //               [x: string]: never;
+    //           };
+    //           status: RunStatus;
+    //           output: string;
+    //       }
```
leeweisberger, line 43: Is it worth it to test this multiple times when it hasn't changed?

packages/airplane/internal/config.test-d.ts
```diff
@@ -1,86 +1,287 @@
...
+  }
+
+  // Async task.
```
leeweisberger, line 46: nice

packages/airplane/internal/config.test-d.ts
```diff
@@ -1,86 +1,287 @@
...
+      },
+      async (_params) => {
+        //   ^? (parameter) _params: {
+        //          shorttext: string;
+        //      }
+      }
+    );
+    type _params = Parameters<typeof myTask>[0];
+    //    ^? type _params = {
+    //           shorttext?: string | undefined;
```
leeweisberger, line 178: noice

leeweisberger (approved):
Let's make sure the views pr gets resolved before we merge this.

I think a few more things can be commented, but tests are really clear!

packages/airplane/internal/builtins/email.ts
colinking: I didn't want to implement support for the map/list parameter types that built-ins use in the type system rn.

packages/airplane/internal/config.ts
colinking: Yep!

packages/airplane/internal/params/params.ts
```diff
@@ -0,0 +1,117 @@
...
+  type: string;
```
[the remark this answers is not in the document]
colinking: Good catch. Fixed.

packages/airplane/internal/params/params_serialized.ts
colinking: It's for consistency, but yeah It's legitimate difference in how serialized parameters look, vs. how they look in the UI. I'm going to leave as-is.

packages/airplane/internal/config.test-d.ts
colinking: Yep, I ran into various bugs that showed up in these types.

PR #203, opened 2022-11-18
Fix additional inline config type issues

colinking:
Fixes a few additional issues related to inline config types, building off of the changes in https://github.com/airplanedev/js-sdk/pull/198:

1. Once a `type` was set, no other fields were shown in intellisense. This was fixed by removing the unnecessary default for `TParams` which was getting set to `Record<string, never>`.
2. If a schedule had required parameters, it would not require that a user set `paramValues`.
3. If `parameters` were set to the literal value of `{}`, the RunFunc's param type would include `| undefined` even though it would always be set. This was fixed by removing the unnecessary conditional from `RunFunc`.
4. The output type had to be set explicitly with `Promise<...>` if the RunFunc was async. This is now optional.

An example showing (1) fixed:

CleanShot 2022-11-18 at 10 58 07

Resolves TSK-634 -- specifically (3), (5), and (7).

cc @ericliuche


colinking:
* main
  * **PR #203** <a href="https://app.graphite.dev/github/pr/airplanedev/node-sdk/203" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/airplanedev/node-sdk/203?utm_source=stack-comment).

packages/airplane/internal/config.test-d.ts
```diff
@@ -290,4 +291,99 @@ const _task = async () => {
...
+        slug: "greet",
+        parameters: {
+          shorttext: "shorttext",
+        },
+      },
+      (_params) => "hello world"
+    );
+    type _myTask2Output = Awaited<ReturnType<typeof _myTask2>>["output"];
+    //   ^? type _myTask2Output = string
+    const _myTask3 = task<InputParams, string>(
```
leeweisberger, line 319: What's the diff between this and _myTask1?

packages/airplane/internal/config.test-d.ts
```diff
@@ -290,4 +291,99 @@ const _task = async () => {
...
+        slug: "greet",
+        parameters: {
+          shorttext: "shorttext",
+        },
+      },
+      async (_params) => "hello world"
+    );
+    type _myTask3Output = Awaited<ReturnType<typeof _myTask3>>["output"];
+    //   ^? type _myTask3Output = string
+    const _myTask4 = task<InputParams, Promise<string>>(
```
leeweisberger, line 330: And this and _myTask2?

packages/airplane/internal/config.ts
```diff
@@ -58,24 +57,36 @@ export type WrappedRunFunc<
       executeOptions?: ClientOptions
     ) => Promise<Run<Resolve<InputToSerializedParamValues<TParams>>, Awaited<TOutput>>>;
 
-export const task = <
-  TParams extends InputParams = Record<string, never>,
-  // eslint-disable-next-line @typescript-eslint/no-explicit-any
-  TOutput = any
->(
+/**
+ * DefaultEmpty is a helper that defaults `TParams` to an empty object if it detects
```
leeweisberger, line 61: I do not understand what this is doing. Is there an example that might make it clearer?

leeweisberger (approved):

packages/airplane/internal/config.test-d.ts
colinking: This is an `async` function. Added comments.

packages/airplane/internal/config.ts
colinking: Yeah, I had trouble explaining what this does. I rewrote this comment with an example which should make it much clearer. Let me know if that helped!

PR #206, opened 2022-11-21
Add prompt tuples

ericliuche:
This change adds a new `prompt` function inside `packages/airplane/internal/prompt/index.ts` that supports the same parameter format as inline task configuration. Doing so allows us to support prompt tuples


packages/airplane/internal/params/params_output.ts
```diff
@@ -11,3 +14,35 @@ export type OutputParamValueByType = Omit<
...
+  : Optional<
+      {
+        [TSlug in keyof TParams]: OutputParamValueByType[NormalizeAsObject<TParams[TSlug]>["type"]];
+      },
+      OptionalParamSlugsWithoutDefaults<NormalizeAsObjects<TParams>>
+    >;
+
+export const deserializeParams = <TParams extends InputParams>(
+  params: TParams,
+  paramValues: InputToOutputParamValues<TParams>
```
colinking, line 32:
```suggestion
  paramValues: InputToSerializedParamValues<TParams>
```

packages/airplane/internal/params/params_serialized.ts
```diff
@@ -61,7 +68,106 @@ export const serializeParamValue = (paramValue: InputParamValue): SerializedPara
     return paramValue.id;
   }
   if (paramValue instanceof Date) {
-    return toISOString(paramValue);
+    // TODO: Currently we incorrectly serialize `date` parameters the same as `datetime`.
+    //       `date` parameters need to instead be formatted as `YYYY-MM-DD`.
+    return paramValue.toISOString();
```
ericliuche, line 73: this was incorrect before. `toISOString(paramValue)` created a format our API doesn't support

packages/airplane/internal/params/params_serialized.ts
```diff
@@ -61,7 +68,106 @@ export const serializeParamValue = (paramValue: InputParamValue): SerializedPara
...
+        type: type,
+        component: component,
+        desc: param.description && dedent(param.description),
+        default: serializeParamValue(param.default),
+        constraints: {
+          optional: optional,
+          options:
+            "options" in param && param.options != null
+              ? // eslint-disable-next-line @typescript-eslint/no-explicit-any
+                serializeOptions(param.options as any, slug, optional)
```
ericliuche, line 128: i wasn't able to figure out how to get this type to work :(

packages/airplane/internal/params/params_serialized.ts
```diff
@@ -61,7 +68,106 @@ export const serializeParamValue = (paramValue: InputParamValue): SerializedPara
...
+              : undefined,
+          regex,
+        },
+      });
+    }
+  }
+  return serializedParams;
+};
+
+const serializeOptions = <TOption extends InputParamValueByType[keyof InputParamValueByType]>(
```
ericliuche, line 138: before, the option values were represented as the index into the options array, why was this the case @colinking? it created odd artifacts where if you submit a prompt with options in slack, it would render the option values as the index value

packages/airplane/internal/params/params_output.ts
```diff
@@ -11,3 +17,22 @@ export type OutputParamValueByType = Omit<
   date: Date;
   datetime: Date;
 };
+
+export const deserializeParamValues = <TParams extends InputParams>(
+  params: TParams,
+  paramValues: InputToSerializedParamValues<TParams>
+): Resolve<InputToOutputParamValues<TParams>> => {
+  // eslint-disable-next-line @typescript-eslint/no-explicit-any
+  const output = {} as Record<keyof TParams, any>;
```
ericliuche, line 26: wasn't able to get this type to work without `any`

packages/airplane/internal/prompt/index.test.ts
```diff
@@ -7,6 +7,7 @@ describe("select", () => {
     const unregister = registerRuntime(
       "mock" as RuntimeKind,
       createMockRuntime({
+        // @ts-ignore
```
ericliuche, line 10: wasn't able to get this type working

packages/airplane/internal/prompt/index.test.ts
```diff
@@ -60,6 +61,7 @@ describe("select", () => {
...
```
ericliuche, line 64: wasn't able to get this type working

packages/airplane/internal/params/params_serialized.ts
colinking: It was an easy way to get a unique identifier of each element. e.g. if you have a list of objects, do you have to serialize the object to get an identifier?

colinking:
@ericliuche is this ready for another review? Just re-request from me

packages/airplane/internal/params/params_serialized.ts
ericliuche: our parameter definition format doesn't accept ISO strings, trying to define a inline task date param or prompt param with options/defaults surfaces
`Error: Api: 400 - Parameters[1].default: Dates must be formatted as '2006-01-02'`

colinking: Ah, that's a bug then. Here's the PR: https://github.com/airplanedev/airport/pull/3503/files

packages/airplane/internal/prompt/index.ts
```diff
@@ -209,3 +209,18 @@ export const select = async <TOption = unknown>(
 
   return options[i];
 };
+
+export const prompt = async <TParams extends PromptInputParams = Record<string, never>>(
```
ericliuche, line 213: i wasn't able to get `airplane.prompt(...)` and `airplane.prompt.text(...)` both to work, i tried something like
```
import { prompt } from "./internal/prompt";
import * as promptFunctions from "./internal/prompt";

prompt.confirm = promptFunctions.confirm;
prompt.text = promptFunctions.text;
prompt.longText = promptFunctions.longText;
prompt.sql = promptFunctions.sql;
prompt.number = promptFunctions.number;
prompt.integer = promptFunctions.integer;
prompt.boolean = promptFunctions.boolean;
prompt.date = promptFunctions.date;
prompt.datetime = promptFunctions.datetime;
prompt.select = promptFunctions.select;
```
but when trying to use `prompt.option.select` i would get 

Screen Shot 2022-11-23 at 10 17 35 AM

ericliuche: and for backwards compat im guessing we'd also keep around
```
packages/airplane/prompt.ts

// Supports:
// import { text } from 'airplane/prompt'
export * from "./internal/prompt";

// Supports:
// import prompt from 'airplane/prompt'
export * as default from "./internal/prompt";
```
for now?

ericliuche:
@colinking ready now!

packages/airplane/internal/execute.test.ts
```diff
@@ -35,7 +35,7 @@ describe("execute", () => {
         paramValues: {
           name: "colin",
           file: "testid",
-          date: "2022-10-31T00:00:00-07:00",
```
colinking, line 38: This looks wrong. We purposefully serialize date types in the current user's timezone so that we don't get "off-by-one" bugs where switching to UTC changes the day.

packages/airplane/internal/params/params.ts
```diff
@@ -25,6 +25,28 @@ export type Param = { type: ParamType } & (
   | ConfigVarParam
 );
 
+// TODO: support configvars and uploads
```
colinking, line 28: Can you add a Linear? Easier to track.

packages/airplane/internal/params/params_input.ts
```diff
@@ -54,6 +62,13 @@ type InputParam = ParamType | Param;
  */
 export type InputParams = Record<string, InputParam>;
 
+/**
+ * PromptInputParams are a set of parameters keyed by parameter slug available to prompts.
+ *
+ * The order of parameters within the object is retained for ordering parameters in form UIs.
+ */
```
colinking, line 69:
```suggestion
/**
 * PromptInputParams are the subset of InputParams that are available to prompts.
 *
 * Currently config variables and files are not supported by prompts. Once those are supported,
 * we can remove this in favor of using InputParams directly.
 */
```

packages/airplane/internal/params/params_output.test.ts
```diff
@@ -0,0 +1,87 @@
+import { deserializeParamValues } from "./params_output";
+
+describe("deserializeParamValues", () => {
```
colinking, line 3: nice

packages/workflow-runtime/internal/index.ts
```diff
@@ -92,20 +93,22 @@ export const runtime: RuntimeInterface = {
     throw new Error("Executing runbooks from workflows is not supported.");
   },
 
-  prompt: async (prompt: PromptConfig): Promise<SerializedParamValues> => {
+  prompt: async <TParams extends InputParams = Record<string, never>>(
```
colinking, line 96: I find it easier to avoid defaults unless there's a good reason for them (e.g. a type exported to end-users that won't get inferred).

packages/workflow-runtime/internal/index.ts
```diff
@@ -92,20 +93,22 @@ export const runtime: RuntimeInterface = {
...
+    prompt: PromptConfig
```
colinking, line 97: nit: rename to `CreatePromptRequest`. Better conveys this is an API type. I thought it was the user-facing type at first.

packages/airplane/internal/prompt/index.ts
```diff
@@ -27,7 +31,7 @@ export const text = async (name: string, opts: PromptOptions<string> = {}): Prom
     cancelText: opts.cancelText,
     confirmText: opts.confirmText,
   });
-  return values[param.slug] as string;
```
colinking, line 30: :awyis: I love this

packages/airplane/internal/prompt/index.ts
```diff
@@ -111,10 +115,8 @@ export const date = async (
...
     confirmText: opts.confirmText,
   });
-  const d = values[param.slug] as string;
-
   // Convert back to Dates before passing back.
-  return new Date(d);
+  return new Date(values[param.slug]);
 };
 
 export const datetime = async (
```
colinking, line 122: Minor, but would it make sense to rewrite all these methods to just call `.prompt()`? e.g. you wouldn't need to do the conversion to output values.

packages/airplane/internal/prompt/index.ts
colinking: Hmm, weird. I'll TAL.

colinking (commented):
Oof there's a lot here 😄 First pass of comments.

packages/airplane/internal/params/params_serialized.ts
colinking: Hmm, for some reason it's tripping on the union types here:

CleanShot 2022-11-23 at 12 59 06

And reducing that to just `string | Date`:

CleanShot 2022-11-23 at 12 59 34

Not sure the fix, can leave as-is.

packages/airplane/internal/params/params_serialized.ts
```diff
@@ -61,7 +68,106 @@ export const serializeParamValue = (paramValue: InputParamValue): SerializedPara
...
+  options: Array<TOption | SelectOption<TOption>>,
+  slug: string,
+  optional: boolean
+): Array<{ label: string; value: SerializedParamValue }> => {
+  if (options.length === 0 && !optional) {
+    throw new Error(`The required prompt parameter, "${slug}", has no options to select from.`);
```
colinking, line 144: This function isn't directly related to prompts.

```suggestion
    throw new Error(`The required parameter, "${slug}", has no options to select from.`);
```

colinking: Hmm, does this match our API semantics? I would almost assume if a user passes an empty array of options, we would ignore it.

packages/airplane/internal/params/params_serialized.ts
```diff
@@ -61,7 +68,106 @@ export const serializeParamValue = (paramValue: InputParamValue): SerializedPara
...
+      //       `date` parameters need to instead be formatted as `YYYY-MM-DD`.
+      return option.toISOString();
+    }
+    return JSON.stringify(option);
+  };
+
+  return options.map((option) => ({
+    label: optionToLabel(option),
+    value: serializeParamValue(
+      typeof option === "object" && "value" in option ? option.value : option
```
colinking, line 170: I'd add a helper for this:

```
const isOptionObject = <TOption>(
  opt: TOption | SelectOption<TOption>
): opt is SelectOption<TOption> => {
  return typeof opt === "object" && opt != null && "value" in opt;
};
```

FYI I noticed `label` is required in `SelectOption` but should be optional. Can you fix?

packages/airplane/internal/params/params_serialized.ts
```diff
@@ -61,7 +68,106 @@ export const serializeParamValue = (paramValue: InputParamValue): SerializedPara
...
+      case "boolean":
+      case "number":
+        return String(option);
+      case "string":
+        return option;
+    }
+    if (option instanceof Date) {
+      // TODO: Currently we incorrectly serialize `date` parameters the same as `datetime`.
+      //       `date` parameters need to instead be formatted as `YYYY-MM-DD`.
+      return option.toISOString();
```
colinking, line 162: Same here as above -- let's continue to use `toISOString(option)` so we retain the time zone when serialized for dates.

packages/airplane/internal/prompt/index.ts
colinking, line 213: You'll want to remove this `Record<string, never>` default as it breaks intellisense:

before:
CleanShot 2022-11-23 at 13 15 28

after:

CleanShot 2022-11-23 at 13 15 47

packages/airplane/internal/prompt/index.ts
colinking: Doing so should fix https://linear.app/airplane/issue/TSK-465/prompt-output-should-be-undefined-when-optional for these methods

packages/airplane/internal/prompt/index.ts
colinking: Seems like you can just do:

```
export const prompt = async <TParams extends PromptInputParams = Record<string, never>>(
  params: TParams,
  opts: BasePromptOptions = {}
): Promise<Resolve<InputToOutputParamValues<TParams>>> => {
  // ...
};
prompt.confirm = confirm;
prompt.integer = integer;
// ...
```

colinking (changes requested):
LGTM, apart from some minor comments below. Is the plan to keep the old methods, or do we want to deprecate them? What's the plan around adding the alternative array notation?

colinking:
@ericliuche one more ask -- can you add JSDoc comments to the prompt methods, similar to what I did in this PR? https://github.com/airplanedev/js-sdk/pull/208

packages/airplane/internal/params/params_serialized.ts
```diff
@@ -61,7 +68,106 @@ export const serializeParamValue = (paramValue: InputParamValue): SerializedPara
...
+          optional: false,
+        },
+      });
+    } else {
+      const [type, component] = serializeParamType(param.type);
+      const regex = "regex" in param ? param.regex : undefined;
+      const optional = param.required === false ? true : false;
+      serializedParams.push({
+        slug: slug,
+        name: param.name ? param.name : slug,
```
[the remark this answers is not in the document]
ericliuche: i'll add a task to do this in a later change, we already do it for the python sdk
`name=param_config.name or inflection.humanize(default_slug),`

packages/airplane/internal/params/params_serialized.ts
ericliuche: yup you're right. should we add the check to parameter verification?

Screen Shot 2022-11-28 at 3 40 56 PM

packages/airplane/internal/prompt/index.ts
ericliuche: hmm would that require us having to write a template that converts `PromptOptions` to `InputParams`?

ericliuche:
`Is the plan to keep the old methods, or do we want to deprecate them?`
i think we should deprecate the old methods, should i just mark them as deprecated here?

`What's the plan around adding the alternative array notation?`
we can do that in a follow up change, i'll add a task for it

colinking:
> i think we should deprecate the old methods, should i just mark them as deprecated here?

Sweet, use this annotation for that and VSCode will mark usage with a strikethrough: https://jsdoc.app/tags-deprecated.html

packages/airplane/internal/params/params_serialized.ts
colinking: We apparently do this in our UI:

CleanShot 2022-11-28 at 18 05 42

Returning an error from the API SGTM. Can you add a Linear for that?

ericliuche: https://linear.app/airplane/issue/TSK-1193/add-parameter-verification-check-that-empty-options-is-not-allowed

ericliuche:
`JSDoc comments to the prompt methods` 

@colinking i just added docs to the new `prompt` method I added since the other methods are deprecated now. is that fine?

packages/airplane/internal/prompt/index.ts
```diff
@@ -17,128 +24,149 @@ export const confirm = async (opts: BasePromptOptions = {}): Promise<void> => {
...
-
   // Convert back to Dates before passing back.
-  return new Date(d);
+  return new Date(values[param.slug]);
 };
 
 /**
  * Prompts a user to select one item from a list of options.
  *
+ * @deprecated since v0.2.23, use `prompt()` instead
```
colinking, line 169: This method's actually a convenience method (given an arbitrary list of elements, you can create a select parameter from it), so there's actually no direct replacement with `prompt`. Is there a way we can keep it or something similar around?

packages/airplane/internal/prompt/prompt.test-d.ts
```diff
@@ -0,0 +1,95 @@
+import { prompt } from "./index";
+
+const _prompt = async () => {
+  // Prompt with no parameters without arg.
+  {
+    const _myPrompt = prompt({});
```
colinking, line 6: add twoslash check

packages/airplane/internal/params/params_serialized.test.ts
```diff
@@ -0,0 +1,623 @@
...
+        default: "default",
+        options: ["default", "foo"],
+      },
+      d: {
+        type: "shorttext",
+        options: [{ label: "foo", value: "bar" }],
+      },
+      e: {
+        type: "shorttext",
+        regex: "$.*^",
```
colinking, line 23:
```suggestion
        regex: "^.*$",
```

colinking (approved):

packages/airplane/internal/prompt/index.ts
ericliuche: ah good point, will just remove the deprecated tag for now

PR #207, opened 2022-11-22
Bring prompts to standard task runs

colinking:
Enables prompt support in the standard runtime. We'll roll this out to the docs soon.

Resolves TSK-1159


colinking:
* main
  * **PR #207** <a href="https://app.graphite.dev/github/pr/airplanedev/node-sdk/207" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/airplanedev/node-sdk/207?utm_source=stack-comment).

packages/airplane/internal/runtime/dev.ts
```diff
@@ -41,21 +40,7 @@ export const runtime: RuntimeInterface = {
   },
 
   prompt: async (prompt: PromptConfig, opts?: ClientOptions): Promise<SerializedParamValues> => {
```
colinking, line 42: The `dev` runtime now just points at the standard runtime. We could get rid of it, but I'll leave it for now in case it drifts from the standard runtime at some point.

examples/standard/main.airplane.ts
```diff
@@ -9,7 +9,7 @@ export default airplane.task(
         name: "Name",
         slug: "name",
```
ericliuche, line 10: just noticed and off topic, but does this slug parameter have any effect? don't we take the slug from the key

ericliuche (approved):

examples/standard/main.airplane.ts
colinking: Good catch! Removed.

PR #213, opened 2022-12-01
Un-deprecate prompt.confirm

ericliuche:


packages/airplane/internal/prompt/index.ts
```diff
@@ -243,7 +246,8 @@ export const select = async <TOption = unknown>(
  * Each prompt contains a parameter form, similar to what you see when executing a task or runbook.
  * When a prompt is created, the run pauses⁠ and waits for a response⁠ before continuing.
  * By default, anyone who can access the associated run can also respond to the prompt,
- * but this can be restriced by specifying reviewers in the prompt's options.
+ * but this can be restriced by specifying reviewers in the prompt's options. Use `prompt.confirm` to
```
colinking, line 249:
```suggestion
 * but this can be restricted by specifying reviewers in the prompt's options. Use `prompt.confirm` to
```

colinking: This line kinda implies that you can't do no parameters with this method? Or at least it could be read that way. Can you split it out as "As a convenience, you can use `prompt.confirm` to create prompts with no parameters."

packages/airplane/internal/prompt/index.ts
```diff
@@ -259,25 +263,14 @@ export const select = async <TOption = unknown>(
  *     { confirmText: "Accept refund" }
  *   );
  * @example
- *   // Prompt with a list of options to choose from.
+ *   // Prompt with a list of options to choose from. Use `prompt.select` to create a prompt with a list
```
colinking, line 266: Same here -- as a convenience...

colinking (approved):

PR #214, opened 2022-12-02
Add task/view definition converter helper

ericliuche:
Adds a `runtime` field to `task` and adds `deprecated` to `workflows`. Also adds a helper function to convert between configuration and task definition.

Still testing...


packages/airplane/internal/config.ts
```diff
@@ -351,14 +371,100 @@ export const workflow = <TParams extends InputParams, TOutput>(
...
-    config: config,
+    type: "task",
+    config: updatedConfig,
     baseFunc: f,
+    convertToDef: convertToTaskDef(updatedConfig),
   };
   return wrappedF;
 };
 
+type TaskDef<TParams extends InputParams> = {
```
ericliuche, line 385: this mimics `type Definition_0_3 struct {` in `lib`

packages/airplane/internal/config.ts
```diff
@@ -351,14 +371,100 @@ export const workflow = <TParams extends InputParams, TOutput>(
...
+        slug: slug,
+        name: param.name ? param.name : slug,
+        type: param.type,
+        description: param.description,
+        default: serializeParamValue(param.default),
+        required: param.required,
+        options:
+          "options" in param
+            ? // eslint-disable-next-line @typescript-eslint/no-explicit-any
+              serializeOptions(param.options as any, slug, !param.required)
```
ericliuche, line 440: before, we weren't explicitly serializing options and instead relying on JSON.serialize

packages/airplane/internal/params/params.ts
```diff
@@ -149,6 +149,7 @@ type ConfigVarParam = {
    */
   type: "configvar";
 } & ParamBase<InputParamValueByType["configvar"]> &
+  ParamRegex &
```
ericliuche, line 152: configvars are regexable

packages/airplane/internal/config.ts
```diff
@@ -328,6 +347,7 @@ export const task = <TParams extends InputParams, TOutput>(
  * Airplane Workflows are currently in beta. To learn more, see
  * [the Workflow documentation](https://docs.airplane.dev/workflows/overview).
  *
+ * @deprecated since v0.3.0, use `task()` instead and set `runtime: "workflow"`
```
colinking, line 350: Are you planning to release this as `0.3.0`? Any reason not to make this a patch release, since it doesn't include breaking changes?

packages/airplane/internal/config.ts
```diff
@@ -351,14 +371,100 @@ export const workflow = <TParams extends InputParams, TOutput>(
     >;
   };
 
+  const runtime: "workflow" | "task" = "workflow";
```
colinking, line 374:
```suggestion
  const runtime: "workflow" | "standard" = "workflow";
```

colinking: Define this union somewhere, so we can re-use it.

packages/airplane/internal/config.ts
```diff
@@ -351,14 +371,100 @@ export const workflow = <TParams extends InputParams, TOutput>(
...
+  const updatedConfig = { ...config, runtime: runtime };
   wrappedF.__airplane = {
-    type: "workflow",
-    config: config,
+    type: "task",
+    config: updatedConfig,
     baseFunc: f,
+    convertToDef: convertToTaskDef(updatedConfig),
```
colinking, line 380: It might make more sense to export this as a utility function? It doesn't really need to be a property. That would also get rid of the closure, making it a bit easier to read IMO.

```
export generateTaskDefinition(task: WrappedAirplaneTask, fileName: string, entrypointFunc: string) TaskDefinition
export generateViewDefinition(task: WrappedAirplaneView) ViewDefinition
```

colinking: You should then be able to do a conditional `require` in the parser, to handle earlier versions of the JS SDK.

colinking (changes requested):
Left some comments on the convertToDef logic. Can we move that into a separate PR so that we can merge the `runtime` changes?

packages/airplane/internal/config.ts
ericliuche: hmm yeah i guess this is just deprecating workflows and adding a field, ill change to 0.2.6

packages/airplane/internal/config.ts
ericliuche: the reason i avoided this was i didn't want `lib` to import the sdk, which it doesn't right now. we'd then have to specify a version for the sdk in the parser's package.json and it felt weird to potentially have two different versions of the sdk running, one for lib and one for user code

colinking: Hmm, okay. How about exporting a reference to the conversion function, where the conversion function has the same signature as my first comment?

```
{
	convertToDef: generateTaskDefinition
}
```

PR #215, opened 2022-12-02
Support fetching files via AirplaneFile

colinking:
Updates `AirplaneFile` to support fetching file contents. Supports both runtimes.

Resolves TSK-1171

cc @geodn @kellymtrinh @leeweisberger


colinking:
* main
  * **PR #215** <a href="https://app.graphite.dev/github/pr/airplanedev/node-sdk/215" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/airplanedev/node-sdk/215?utm_source=stack-comment).

packages/workflow-runtime/internal/index.ts
```diff
@@ -1,19 +1,20 @@
-import { defineSignal, setHandler, condition, uuid4 } from "@temporalio/workflow";
-import { proxySinks } from "@temporalio/workflow";
+import { TextDecoder, TextEncoder } from "@temporalio/common/lib/encoding";
```
colinking, line 1: Turns out Temporal exposes one of the TextEncoder/TextDecoder polyfills.

packages/workflow-runtime/internal/index.ts
```diff
@@ -157,6 +160,28 @@ const passthroughOptions = (opts: ClientOptions): ClientOptions => {
...
+/**
+ * Attach global polyfills. Unlike the standard runtime which runs on Node.js, the workflow runtime
+ * executes within V8 which does not have access to Web or Node APIs.
+ */
+const registerWorkflowPolyfills = async () => {
+  // Polyfill `TextEncoder` and `TextDecoder`. This is used by our Blob polyfill below. We
+  // use the polyfills exported by Temporal, but they are not standards-compliant so we have to
+  // cast. They seem to work well-enough for our use-case though. Unfortunately, there are just
+  // not many high-quality TextEncoder/TextDecoder polyfills since Node/browsers have such good
+  // coverage nowadays.
```
colinking, line 178: Maybe one day there will be open-sourced implementation of the standards produced by https://wintercg.org that Temporal could adopt. Would make things much easier!

packages/airplane/package.json
```diff
@@ -21,23 +21,23 @@
...
-    "timezoned-date": "^3.0.2",
+    "timezoned-date": "3.0.2",
     "ts-jest": "29.0.3"
   },
   "dependencies": {
     "@azure/abort-controller": "1.1.0",
     "@opentelemetry/api": "1.2.0",
     "cross-fetch": "3.1.5",
     "fast-equals": "4.0.3",
-    "fetch-blob": "3.2.0",
```
colinking, line 40: This was only used in tests. Moved to dev dependencies.

packages/workflow-runtime/internal/index.ts
```diff
@@ -157,6 +160,28 @@ const passthroughOptions = (opts: ClientOptions): ClientOptions => {
...
   registerRuntime("workflow", runtime);
+
+  await registerWorkflowPolyfills();
+};
+
+/**
+ * Attach global polyfills. Unlike the standard runtime which runs on Node.js, the workflow runtime
+ * executes within V8 which does not have access to Web or Node APIs.
+ */
+const registerWorkflowPolyfills = async () => {
```
ericliuche, line 173: qq: why does this have to be async?

packages/airplane/internal/file/file.ts
```diff
@@ -1,25 +1,193 @@
...
+   */
+  static #extractName(url: string): string {
+    if (!url) {
+      return "";
+    }
+    // Extract the pathname from the URL. Note we can't use the URL global in the
+    // workflows runtime.
+    //
+    // Remove the protocol.
+    const u = url.replace(/^[a-z]+:\/\//, "");
```
ericliuche, line 176: can you not use `URL().pathname` because temporal <> node?

packages/airplane/internal/file/file.ts
```diff
@@ -1,25 +1,193 @@
...
+  // eslint-disable-next-line @typescript-eslint/no-explicit-any
+  constructor(...args: any[]) {
+    if (isAirplaneFileObject(args[0])) {
+      const o = args[0];
+      this.id = o.id;
+      this.url = o.url;
+      this.name = AirplaneFile.#extractName(o.url);
+      this.size = 0;
+      this.type = "";
+    } else if (isBlob(args[0])) {
```
ericliuche, line 91: To double check, since users are able to provide their own blobs, they'd be able to wrap perhaps an s3 url and pass it into another task? I wonder if we'd display that correctly on the frontend.

ericliuche (approved):

packages/workflow-runtime/internal/index.ts
colinking: Good catch! I was previously performing dynamic imports, so I needed to make this an async function. Fixed.

packages/airplane/internal/file/file.ts
colinking: Temporal workflows execute in V8, not Node. We would need to add a `URL` polyfill for that to work.

packages/airplane/internal/file/file.ts
colinking: We don't support uploading yet, but I'm going to work on that next. We need that for file displays and for passing `AirplaneFile`'s that were created locally as parameters into other tasks.

PR #216, opened 2022-12-02
Adds runtime field to task definition

ericliuche:
Also deprecates workflow definition and fixes configvar regex type


packages/airplane/internal/config.ts
```diff
@@ -328,6 +341,7 @@ export const task = <TParams extends InputParams, TOutput>(
  * Airplane Workflows are currently in beta. To learn more, see
  * [the Workflow documentation](https://docs.airplane.dev/workflows/overview).
  *
+ * @deprecated since v0.2.6, use `task()` instead and set `runtime: "workflow"`
```
colinking, line 344:
```suggestion
 * @deprecated since v0.2.23, use `task()` instead and set `runtime: "workflow"`
```

packages/airplane/internal/config.ts
```diff
@@ -351,9 +365,11 @@ export const workflow = <TParams extends InputParams, TOutput>(
     >;
   };
 
+  const runtime: TaskRuntime = "workflow";
+  const updatedConfig = { ...config, runtime: runtime };
   wrappedF.__airplane = {
-    type: "workflow",
-    config: config,
+    type: "task",
```
colinking, line 371: this requires a lib change to always look at `config.runtime` -- maybe leave as `workflow` for now unless you plan to merge a `lib` change first?

colinking (approved):

packages/airplane/internal/config.ts
```diff
@@ -317,7 +330,7 @@ export const task = <TParams extends InputParams, TOutput>(
   };
 
   wrappedF.__airplane = {
-    type: "task",
+    type: config.runtime === "workflow" ? "workflow" : "task",
```
ericliuche, line 333: @colinking also added this, does it look right to you?

PR #220, opened 2022-12-02
Add prompt.background

ericliuche:
Adds support for `promptBackground` which returns a `Prompt` object without waiting for user input. Input can then be waited on by calling `await prompt.wait()`. Also adds `confirmBackground` and `selectBackground` for convenience.


packages/airplane/internal/prompt/index.ts
```diff
@@ -15,7 +15,18 @@ export type { PromptOptions };
...
+};
+
+/**
+ * Prompts a user for confirmation without requiring parameters and does not wait for confirmation.
+ * Wait on the confirmation by calling `.wait()` on the returned object. As a convenience,
+ * you can use `confirm()` to create a prompt and wait for a response.
+ *
+ * @param opts Configuration options.
+ */
+export const confirmBackground = async (opts: BasePromptOptions = {}): Promise<PromptConfirm> => {
```
colinking, line 28: AD, let's remove the confirm/select background methods.

packages/airplane/internal/prompt/index.ts
```diff
@@ -276,7 +322,48 @@ export const prompt = async <TParams extends PromptInputParams>(
   params: TParams,
   opts: BasePromptOptions = {}
 ): Promise<Resolve<InputToOutputParamValues<TParams>>> => {
-  const values = await getRuntime().prompt<TParams>({
+  return await (await promptBackground<TParams>(params, opts)).wait();
+};
+
+/**
+ * Prompts an operator for input and does not wait for the prompt to be sumbmitted. The prompt response
```
colinking, line 329:
```suggestion
 * Prompts an operator for input and does not wait for the prompt to be submitted. The prompt response
```

packages/airplane/internal/prompt/index.ts
```diff
@@ -276,7 +322,48 @@ export const prompt = async <TParams extends PromptInputParams>(
...
+ * can be retrieved by calling `.wait()` on the returned object. As a convenience,
+ * you can use `prompt()` to create a prompt and wait for a response.
```
colinking, line 331: If we made this `airplane.prompt.background()`, that makes it clear that `.prompt()` is the recommended way of calling prompts (since a user who looks at the hints at `airplane.` won't see both `prompt` and `promptBackground`). Let's do that unless you have opinions otherwise?

packages/airplane/internal/prompt/index.ts
```diff
@@ -285,11 +372,50 @@ export const prompt = async <TParams extends PromptInputParams>(
     confirmText: opts.confirmText,
     description: opts.description,
   });
-  return deserializeParamValues<TParams>(params, values);
+  return new Prompt<TParams>(params, promptID);
 };
 
+export class Prompt<TParams extends PromptInputParams> {
+  constructor(public readonly params: TParams, readonly promptID: string) {
```
colinking, line 379: `public` is the default, so you don't need to set it.

packages/airplane/internal/prompt/index.ts
```diff
@@ -285,11 +372,50 @@ export const prompt = async <TParams extends PromptInputParams>(
...
+    this.params = params;
+    this.promptID = promptID;
+  }
+  async wait(): Promise<Resolve<InputToOutputParamValues<TParams>>> {
```
colinking, line 383: Add doc comment.

colinking, line 383: You shouldn't need the `Resolve<>`.

packages/airplane/internal/runtime/standard.ts
```diff
@@ -94,14 +94,22 @@ export const runtime: RuntimeInterface = {
   prompt: async <TParams extends InputParams = Record<string, never>>(
```
colinking, line 94: Since `prompt` just calls the other two methods, let's remove it to keep the `runtime` API slim.

colinking (changes requested):
LGTM!

packages/airplane/internal/prompt/index.ts
ericliuche: without it there's a bunch of unwanted wrapping
```
+    //     ^? const _myResult: Optional<{
     //            shorttext: string;
     //            longtext: string;
     //            sql: string;
+    //            boolean: boolean;
     //            integer: number;
     //            float: number;
     //            date: Date;
     //            datetime: Date;
-    //        }
+    //        }, never>
```

colinking: Derp, I mixed this up with https://www.typescriptlang.org/docs/handbook/utility-types.html#awaitedtype -- nm!

PR #222, opened 2022-12-04
Fix file parameters: rename to upload

colinking:
The file parameter is called a "File" in our UI and docs, but an "Upload" in our API and task definition format. We would like to rename it to "File" everywhere.

Currently, file parameters do not work with inline config since the type system expects "file" but never converts to "upload" before passing to the API. Users could get file parameters to work by passing "upload" instead and casting to avoid type errors. This PR fixes the types on inline config to line up with this behavior, by expecting "upload" instead of "file".


colinking:
* main
  * **PR #222** <a href="https://app.graphite.dev/github/pr/airplanedev/node-sdk/222" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/airplanedev/node-sdk/222?utm_source=stack-comment).

packages/airplane/internal/config.test-d.ts
```diff
@@ -63,66 +63,270 @@ const _task = async () => {
     //       }
   }
 
-  // Task with all parameters.
+  // Task with each parameter: shorttext.
```
colinking, line 66: TypeScript has an internal limit on the length (in characters) of a type. When it hits that limit, it starts truncating the type by e.g. replacing union types with `... 1 more ...`. For whatever reason, this change caused us to hit that limit in this test so I had to split up each of the parameters into their own tests, so each type is smaller. The types are still equivalent.

colinking:
> Did you change your mind on the comment in: [airplanedev/docs#542](https://github.com/airplanedev/docs/pull/542)? Are we going with upload in the js sdk and file in the ui?

I would still like to make that change, but in the meantime we should get file parameters working.

leeweisberger (approved):

PR #224, opened 2022-12-05
Create run function with deserialized params

leeweisberger:
Create `run` function that we will call from the task shim (instead of basefunc) that deserializes params before passing them in through the task.

In a follow up, we will change the shim to call the `run` function.

cc @ericliuche


packages/airplane/internal/config.ts
```diff
@@ -365,14 +374,33 @@ export const workflow = <TParams extends InputParams, TOutput>(
     >;
   };
 
+  const run = getRunFunc(f, config.parameters);
+
   wrappedF.__airplane = {
     type: "workflow",
     config: config,
     baseFunc: f,
+    run,
```
colinking, line 383: The naming here seems quite confusing -- wrapped function (what a user calls to execute an inline task; expects InputParamValues), baseFunc (the old shim entrypoint; expects SerializedParamValues); run (the new shim entrypoint; expects SerializedParamValues).

Can we export `run` as `baseFunc` instead?

colinking (approved):

packages/airplane/internal/config.ts
leeweisberger: Two reasons I did this
1. For backwards compat. If you upgrade the sdk version but aren't on the latest cli, your task will start breaking. I don't really want to break more than we have to or try to time this perfectly.
2. It's the same convention as python https://github.com/airplanedev/lib/blob/main/pkg/build/universal-python-shim.py#L41

leeweisberger: Doh nevermind I was thinking about why I left baseFunc around...

Ok I can update it we'll just be inconsistent with python and node. This way i don't have to change the shim

colinking: Great! We should update Python at some point, too. cc @ericliuche

PR #233, opened 2022-12-09
Accept File in AirplaneFile constructor

colinking:
From testing Views locally with the latest commit of JS SDK, I noticed we weren't handling `File` types correctly. The `isBlob` check wasn't passing for `File` (which extends `Blob`).

Resolves TSK-1280


colinking:
* main
  * **PR #233** <a href="https://app.graphite.dev/github/pr/airplanedev/node-sdk/233" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/airplanedev/node-sdk/233?utm_source=stack-comment).

packages/airplane/internal/file/file.ts
```diff
@@ -185,9 +185,20 @@ export class AirplaneFile {
...
  * @see https://developer.mozilla.org/en-US/docs/Web/API/Blob
  */
 const isBlob = (b: unknown): b is Blob => {
-  return b != null && typeof b === "object" && b.constructor.name === "Blob";
+  if (b == null || typeof b !== "object") {
+    return false;
+  }
+
+  // Fallback to checking the object's StringTag to handle cases where Blob is not polyfilled.
+  if (("Blob" in globalThis && b instanceof globalThis.Blob) || String(b) === "[object Blob]") {
```
geodn, line 197: I think the test passes with `b.constructor.name === "Blob"` as well as `String(b) === "[object Blob]"`. Any reason to prefer one over the other?

geodn (approved):

packages/airplane/internal/file/file.ts
colinking: The former didn't pass with `fetch-blob`. I believe their constructor name is `_Blob`. The latter checks `Symbol.StringTag` which should be more reliable!

colinking: The latter checks `Symbol.StringTag` which should be more reliable than the constructor name.

PR #241, opened 2022-12-15
Add sleep SDK

kellymtrinh:
Adds the sleep SDK. The standard one uses JS setTimeout.

Usage examples:
`await airplane.sleep("5s")` - we type the duration input to be "123unit" or "123 unit". 

I'm using the `vercel/ms` library since it seems lightweight and is what temporal uses. Note that the latest version is called canary even though it's[ been out for over a year (it's the only one that has TS support](https://github.com/vercel/ms/issues/184)). 

 
# Test Plan
- [x] Deployed a workflow locally  that uses sleep and verified in the temporal UI that a timer started with the correct duration. Checked the run_timers table for correct dates. Verified with print statements that it waited until the end to continue when I awaited the airplane.sleep(). The activity start time and end time calculated off of that looks correct.
- [x] Deployed a standard task locally that uses sleep and verified it waited the specified number of seconds until continuing. (This is implemented with setTimeout)
- [x] Checked error case where I input a bad ms formatted string (ie. "5blah") and checked it returned `airplane_output_set:error "invalid sleep duration"` in the task/workflow.
- [x] Set a duration that's too long and checked the API returns an err message about the duration exceeding the run limit.


packages/workflow-runtime/internal/activities.ts
```diff
@@ -72,6 +77,26 @@ export const createActivities = () => ({
       throw serializeAsTemporalError(err);
     }
   },
+
+  createTimerActivity: async (d: SleepDuration, opts: ClientOptions = {}): Promise<void> => {
+    const client = new Client(opts);
+    try {
+      // Use the time the activity was scheduled as the timer start time
+      const activityStartTime = Context.current().info.scheduledTimestampMs;
```
colinking, line 85: Hmm, it's not clear from their docs if this is when the _current_ activity task was scheduled or when the _first_ activity task was scheduled. Can you check with the Temporal folks?

IOW, is it the first or second "ActivityTaskScheduled" in this diagram: https://docs.temporal.io/concepts/what-is-a-schedule-to-start-timeout

Ideally, we use the first!

colinking (changes requested):

packages/workflow-runtime/internal/activities.ts
kellymtrinh: I'll check with them- 🤔 the JS docs suggest it's the "current" activity everywhere [though](https://github.com/temporalio/sdk-typescript/blob/main/packages/activity/src/index.ts#L113-L153)

kellymtrinh:
(Pending clarification from temporal about activity scheduled time- I'm going to test out logging activity scheduled times for a few activities in the same workflow and see if they're different though. )

packages/airplane/internal/runtime/index.ts
```diff
@@ -78,3 +80,8 @@ export const getRuntime = (): RuntimeInterface => {
 
   return standardRuntime;
 };
+
+export type SleepDuration = {
+  str: string;
+  ms: number;
```
colinking, line 86: Do we still use `str`? Can we just pass through `ms` instead? IOW, `runtime.sleep(ms: number)`

packages/airplane/internal/sleep/index.ts
```diff
@@ -0,0 +1,14 @@
+import ms, { StringValue } from "ms";
+
+import { getRuntime } from "../runtime";
+
+export const sleep = async (duration: StringValue) => {
+  if (!duration) {
+    throw new Error("sleep duration is required");
+  }
+  const durationMs = ms(duration);
+  if (!durationMs) {
```
colinking, line 10: From what I can tell, `ms` will throw errors if the duration is invalid. I skimmed the code and see that it returns `NaN` in some cases and can return `0`. Is that what you are handling here? If so, the error message could be better.

packages/airplane/internal/sleep/index.ts
```diff
@@ -0,0 +1,14 @@
+import ms, { StringValue } from "ms";
+
+import { getRuntime } from "../runtime";
+
+export const sleep = async (duration: StringValue) => {
+  if (!duration) {
+    throw new Error("sleep duration is required");
```
colinking, line 7: This error message seems to make more sense up here?

```suggestion
    throw new Error("expected a duration for how long to sleep (e.g. '1h' to sleep for an hour)");
```

packages/workflow-runtime/internal/activities.ts
```diff
@@ -72,6 +78,25 @@ export const createActivities = () => ({
       throw serializeAsTemporalError(err);
     }
   },
+
+  createSleepActivity: async (d: SleepDuration, opts: ClientOptions = {}): Promise<void> => {
+    const client = new Client(opts);
+    try {
+      // Use the time the activity was scheduled as the timer start time
```
colinking, line 85: Let's dump some context as a comment here. Thanks for checking w/ them in Slack!

```suggestion
      // Compute when the sleep should end based on the timestamp of when this activity
      // was originally scheduled (IOW, if Temporal performs multiple attempts at executing this
      // activity, this timestamp will always be the same).
```

packages/airplane/internal/sleep/index.ts
```diff
@@ -0,0 +1,14 @@
+import ms, { StringValue } from "ms";
+
+import { getRuntime } from "../runtime";
+
+export const sleep = async (duration: StringValue) => {
```
colinking, line 5: Add JSDoc. Also indicate that this method is in beta and may change. See prompts for JSDocs examples.

colinking (approved):

packages/workflow-runtime/internal/activities.ts
kellymtrinh: Screenshot 2022-12-21 at 11 53 29 AM

PR #248, opened 2022-12-23
Add sleep SDK

kellymtrinh:
This is the same as https://github.com/airplanedev/js-sdk/pull/241 but will not merge this until more of the backend functionality (studio, workflows) is there.

Adds the sleep SDK. The standard one uses JS setTimeout.

Usage examples:
`await airplane.sleep("5s")` - we type the duration input to be "123unit" or "123 unit". 

I'm using the `vercel/ms` library since it seems lightweight and is what temporal uses. Note that the latest version is called canary even though it's[ been out for over a year (it's the only one that has TS support](https://github.com/vercel/ms/issues/184)). 

 
# Test Plan
- [x] Deployed a workflow locally  that uses sleep and verified in the temporal UI that a timer started with the correct duration. Checked the run_timers table for correct dates. Verified with print statements that it waited until the end to continue when I awaited the airplane.sleep(). The activity start time and end time calculated off of that looks correct.
- [x] Deployed a standard task locally that uses sleep and verified it waited the specified number of seconds until continuing. (This is implemented with setTimeout)
- [x] Checked error case where I input a bad ms formatted string (ie. "5blah") and checked it returned `airplane_output_set:error "invalid sleep duration"` in the task/workflow.
- [x] Set a duration that's too long and checked the API returns an err message about the duration exceeding the run limit.


kellymtrinh:
(Not merging until later)

colinking:
@kellymtrinh something's up with this PR -- there's no content.

colinking (changes requested):
(since it's empty)

kellymtrinh:
Hmm I think restoring from a deleted branch and trying to recreate the PR didn't work - will open a new one.

PR #252, opened 2023-01-05
Apply fetcher timeouts per request

colinking:
This PR refactors the Fetcher implementation in order to:
1. Apply request timeouts on each request attempt (as opposed to the cumulative time taken across retries)
2. Ensure the response has the proper `Content-Type` header before deserializing a response body as JSON.
3. Improve the retry policy by introducing exponential backoff with jitter.
4. Respect the `Retry-After` headers when computing the retry policy.
5. Respect the `abortSignal` when pausing between retry attempts.
6. Tweak the tests for some minor speed-ups.
7. Increase the maximum number of attempts to 10 from 5.
8. Remove `RequestTimeoutError` by merging it with `HTTPError` which now will log some additional metadata when our shim does `console.error(err)`.
9. Add teamID and runID to `User-Agent`, based [on this discussion](https://airplanedev.slack.com/archives/C032C7W0T5H/p1673022602814219).

This PR also increases the Temporal activity timeouts to accommodate how API requests can take longer than before.

Resolves TSK-1326 - cc @yolken-airplane 

Tested by deploying a standard and a workflow run to stage and ensuring they were able to make API calls as expected.


colinking:
* main
  * **PR #252** <a href="https://app.graphite.dev/github/pr/airplanedev/node-sdk/252" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/airplanedev/node-sdk/252?utm_source=stack-comment).

.vscode/settings.json
```diff
@@ -1,3 +1,5 @@
 {
-  "typescript.tsdk": "node_modules/typescript/lib"
+  "typescript.tsdk": "node_modules/typescript/lib",
+  "jest.jestCommandLine": "yarn workspace airplane run test",
+  "jest.autoRun": "off"
```
colinking, line 4: The jest extension is actually quite nice! This only works for the `airplane` package tests because we'll need to move the jest config file to the root first.

packages/airplane/internal/api/fetcher.ts
```diff
@@ -317,8 +262,196 @@ export class HTTPError extends Error {
...
+
+  const durationSec = parseInt(header, 10);
+  if (!isNaN(durationSec)) {
+    // Convert to milliseconds.
+    return durationSec * 1000;
+  }
+
+  const durationMs = new Date(header).getTime();
+  // Confirm the date is valid before returning.
+  return isNaN(durationMs) ? undefined : durationMs;
```
ericliuche, line 452: have to subtract the current time

packages/airplane/internal/api/fetcher.ts
```diff
@@ -317,8 +262,196 @@ export class HTTPError extends Error {
...
+    controller.signal.addEventListener("abort", onDone);
+    if (controller.signal.aborted) {
+      onDone();
+    }
+  });
+};
+
+const isJSONResponse = (response: Response): boolean => {
+  const contentType = response.headers.get("content-type") ?? "";
+  return contentType.indexOf("application/json") > -1;
```
ericliuche, line 429: `response.headers.get("content-type")?.includes("application/json") ?? false;`?

ericliuche (approved):

packages/airplane/internal/api/fetcher.ts
colinking: Good catch, thanks!

packages/airplane/internal/api/fetcher.ts
colinking: I think it's more readable as-is. Not going to change.

colinking:
Removed the `parent_run_not_found` case (https://github.com/airplanedev/js-sdk/pull/252/commits/d072c6333f75ca2a9e7fda16fcc52fdf364d0971) since we fixed the root cause.

PR #254, opened 2023-01-09
Restrict 4xx codes that are retried

colinking:
There are conditions where these status codes could be retried, but in general, they should not be automatically retried. For example, our API has only one case where a 409 should be retried (and about a dozen cases where a 409/423 should not).

We might add a `retry: boolean` field to the error message in the future, which would allow us to retry the former (very rare) case.


colinking:
* main
  * **PR #254** <a href="https://app.graphite.dev/github/pr/airplanedev/node-sdk/254" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/airplanedev/node-sdk/254?utm_source=stack-comment).

ericliuche (approved):

PR #255, opened 2023-01-10
Add skip sleep for local dev

kellymtrinh:
Updates the local dev runtime for sleep to be a promise that resolves if:
- the sleep is skipped OR
- the sleep finishes

It does this by calling `Promise.all()` on both of those. If the sleep finishes, it stops polling to check if the sleep is skipped. If the sleep is skipped, it aborts the timeout.

Depends on this endpoint existing in local server: 
https://github.com/airplanedev/cli/pull/897

## Test plan
- [x] Tested that skipping a sleep stops waiting, and rest of the run proceeds
- [x] Tested the happy path that a sleep finishes and the rest of the run proceeds as expected 

- Verified nothing happens to deployed tasks or workflows. I tested this locally (not in studio)


packages/airplane/internal/sleep_with_cancel.ts
```diff
@@ -0,0 +1,29 @@
+import { AbortController } from "@azure/abort-controller";
+
+/**
+ * sleepWithCancel returns a Promise that resolves after the provided Date or duration (in milliseconds).
+ *
+ * The returned Promise will resolve sooner if the provided AbortController resolves earlier.
+ */
+export const sleepWithCancel = async (until: Date | number, controller: AbortController) => {
```
colinking, line 8: Let's rename this `setTimeoutWithCancel` to clarify it is unrelated to sleep` and move into `airplane/internal/set_timeout.ts` to clarify it is used outside of `internal/api`.

packages/airplane/internal/sleep_with_cancel.test.ts
```diff
@@ -0,0 +1,18 @@
+import { AbortController } from "@azure/abort-controller";
+
+import { sleepWithCancel } from "./sleep_with_cancel";
+
+describe("sleepWithCancel", () => {
```
colinking, line 5: nice

packages/airplane/internal/runtime/dev.ts
```diff
@@ -55,7 +60,40 @@ export const runtime: RuntimeInterface = {
...
+    });
+    let isSleepFinished = false;
+    const controller = new AbortController();
+
+    const localSleep = new Promise<void>((resolve) => {
+      return sleepWithCancel(duration, controller).then(() => {
+        isSleepFinished = true;
+        resolve();
+      });
+    });
```
colinking, line 77: You won't need the `Promise` here. This is equivalent.

```suggestion
    const localSleep = async () => {
      await sleepWithCancel(duration, controller)
      isSleepFinished = true;
    };
```

packages/airplane/internal/runtime/dev.ts
```diff
@@ -55,7 +60,40 @@ export const runtime: RuntimeInterface = {
...
+          // Abort the timeout if the sleep was skipped.
+          controller.abort();
+          return true;
+        }
+        return null;
+      });
+    };
+
+    // Wait for either the sleep to finish or for the sleep to be skipped.
+    await Promise.all([localSleep, checkSkip(sleepID)]);
```
colinking, line 96: Without wrapping in `Promise` above, you'll now need to call the function.

```suggestion
    await Promise.all([localSleep(), checkSkip(sleepID)]);
```

packages/airplane/internal/runtime/dev.ts
```diff
@@ -55,7 +60,40 @@ export const runtime: RuntimeInterface = {
...
   sleep: async (duration: number): Promise<void> => {
-    return standardRuntime.sleep(duration);
+    const client = new Client();
+    const now = new Date();
+    const sleepID = await client.createSleep({
+      durationMs: duration,
+      until: calculateEndTimeISOString(now.getTime(), duration),
+    });
+    let isSleepFinished = false;
+    const controller = new AbortController();
```
colinking, line 70: Add comment.
```suggestion
    // Create an AbortController to cancel the sleep if it is skipped.
    const controller = new AbortController();
```

packages/airplane/internal/runtime/dev.ts
```diff
@@ -55,7 +60,40 @@ export const runtime: RuntimeInterface = {
...
+    const localSleep = new Promise<void>((resolve) => {
+      return sleepWithCancel(duration, controller).then(() => {
+        isSleepFinished = true;
+        resolve();
+      });
+    });
+
+    const checkSkip = async (id: string) => {
+      const poller = new Poller({ delayMs: 500 });
+      return poller.run(async () => {
```
colinking, line 81: The return value is meaningless, so don't return to make that clear.

```suggestion
      await poller.run(async () => {
```

colinking (approved):

PR #262, opened 2023-01-25
Set dev token header when a tunnel token is present

calufornia:
When running Studio in tunnel mode, we need to include an `X-Airplane-Dev-Token` header in all requests, so that the dev server can verify the request was initiated by the same user that started the ngrok tunnel. We incorporate this logic into the fetcher so that task execution requests (and other calls) from Views or JavaScript tasks can properly communicate with the dev server (see https://github.com/airplanedev/cli/pull/914 for more details).

Tested by using the local version of the js sdk in the @airplane/views package, and then using a local version of the views package in a view. Verified that opening a view in Studio worked.


ameyakhare (approved):
can you add a test plan to the description on how you verified this change works properly?

calufornia:
> can you add a test plan to the description on how you verified this change works properly?

Done—also added a unit test.

PR #264, opened 2023-01-26
Default task + param names to be humanized slug

ericliuche:
Before, the default task and parameter names were just the task and parameter slugs. Now, we first humanize the slug for a prettier name


packages/airplane/package.json
```diff
@@ -38,6 +38,7 @@
     "@azure/abort-controller": "1.1.0",
     "@opentelemetry/api": "1.2.0",
     "fast-equals": "4.0.3",
+    "humanize-string": "^3.0.0",
```
colinking, line 41: pin this dep

packages/airplane/internal/config.ts
```diff
@@ -357,7 +358,7 @@ export const workflow = <TParams extends InputParams, TOutput>(
 
   wrappedF.__airplane = {
     type: "workflow",
-    config: dedentConfig(config),
+    config: normalizeConfig(config),
```
colinking, line 361: Add this to Views, too (just for the view name, since there are no parameters yet).

colinking (approved):
Nice!

packages/airplane/internal/config.ts
colinking: Yeah, IMO it should, to match tasks. I doubt that was intentional.

ericliuche: 👍

leeweisberger:
I was just gonna do part this and suddenly it works!

PR #269, opened 2023-01-31
Update fetcher to support additional methods

colinking:
This PR adds support for the other main HTTP methods.

It also updates the type of query params on `.get()` to allow passing through `null/undefined` (used by Views). It also updates the `headers` parameter to match for consistency.


colinking:
* main
  * **PR #269** <a href="https://app.graphite.dev/github/pr/airplanedev/node-sdk/269" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/airplanedev/node-sdk/269?utm_source=stack-comment).

packages/lib/internal/fetcher.ts
```diff
@@ -100,15 +136,18 @@ export class Fetcher {
       }
     }
 
+    const headers: Record<string, string> = {};
+    for (const [k, v] of Object.entries(opts.headers ?? {})) {
+      if (v) {
+        headers[k] = v;
+      }
```
ericliuche, line 143: will this exclude `0`, `""`, etc?

colinking: It excludes the empty string, which is intentional. The type is `string`, so it won't receive numbers.

colinking: Added a comment in code.

ericliuche (approved):

PR #273, opened 2023-01-31
Add params to Views type

stepzhou:
This makes the `airplane.views` properly type the query params prop.

Resolves PDS-1470


packages/airplane/internal/config.ts
```diff
@@ -392,7 +392,11 @@ export type ViewConfig = {
   description?: string;
 };
 
-export type View = () => ReactNode;
+type ViewProps = {
+  params: Record<string, string>;
```
rose-huang, line 396: why is `params` required instead of optional?

rose-huang (changes requested):

rose-huang (approved):

PR #278, opened 2023-02-01
Fix build warning about unknown RequestOptions

colinking:


colinking:
* main
  * **PR #278** <a href="https://app.graphite.dev/github/pr/airplanedev/node-sdk/278" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/airplanedev/node-sdk/278?utm_source=stack-comment).

packages/lib/index.ts
```diff
@@ -1,3 +1,4 @@
-export { Fetcher, RequestOptions } from "./internal/fetcher";
+export { Fetcher } from "./internal/fetcher";
+export type { RequestOptions } from "./internal/fetcher";
```
ericliuche, line 2: could you leave a comment here explaining why it has to be separate

ericliuche (approved):

packages/lib/index.ts
colinking: Eh, I think the `export type` syntax is enough. Would be nice to lint for this, though I think it shouldn't matter and is only a shortcoming in the webpack build process Temporal is using.

PR #287, opened 2023-02-15
Use get-run-when-terminated endpoint for synchronous execute

zhan-airplane:
Description:
So that we don't need to do polling for the run status

Test plan:
Unit tests


packages/airplane/internal/api/client.ts
```diff
@@ -142,6 +142,15 @@ export class Client {
     }>("/v0/runs/get", { id: runID }, { abortSignal: opts.abortSignal });
   }
 
+  async getRunWhenTerminated<P = unknown>(runID: string, opts: { abortSignal?: AbortSignal } = {}) {
+    return this.fetcher.get<{
+      id: string;
+      status: RunStatus;
+      paramValues: P;
+      taskID?: string;
+    }>("/v0/runs/getWhenTerminated", { id: runID }, { abortSignal: opts.abortSignal });
```
colinking, line 151: You will need to implement this in the CLI first: https://github.com/airplanedev/cli/blob/c96996388314c5cc46ba39b53e002c98806a3ffd/pkg/server/apiext/apiext.go#L44

That also means users will need a minimum version of their CLI before using this version of the SDK. Can you wrap this fetcher call with a `try/catch` and detect if an `HTTPError` with a `404` status code is thrown, in which case we can fallback on calling `get` instead? Therefore, this will work for users with older CLIs.

colinking (changes requested):
LGTM after the CLI is updated.

zhan-airplane:
Going to close this first - the external API design might need some work, so will do that before having any external code depend on it (`web`'s using `apiint`, so even though it's already using the endpoint it's still easier to change)

PR #293, opened 2023-02-17
Pass API headers through fetcher

francescaleung:
Expects environment variable `AIRPLANE_API_HEADERS` to be a serialized JSON object of headers to be passed through requests made from fetcher.

Resolves DTS-323.


packages/lib/internal/fetcher.test.ts
```diff
@@ -521,6 +521,7 @@ describe("headers", () => {
     clientVersion: "1",
```
calufornia, line 521: Can you test the override behavior? i.e. that the request headers override the fetcher headers?

calufornia (approved):

colinking (approved):

PR #295, opened 2023-02-21
Add support for creating child run requests

ericliuche:
Adds routes to get task reviewers, get trigger requests, and create trigger requests to support requesting child runs automatically.

Will add support for workflow child run requests in a later change.


colinking (approved):
🔥

PR #300, opened 2023-02-27
Add view token to header as request opts

kellymtrinh:
Pass view token as a header to the API, if specified.

# Test plan
- If there's no view token, header is not sent.


packages/lib/internal/fetcher.ts
```diff
@@ -168,6 +169,9 @@ export class Fetcher {
     if (this.opts.token) {
       headers["X-Airplane-Token"] = this.opts.token;
     }
+    if (this.opts.viewToken) {
+      headers["X-Airplane-View-Token"] = this.opts.viewToken;
```
leeweisberger, line 173: We have to make sure that we add this to our allow listed list of headers in airport

leeweisberger (approved):

PR #302, opened 2023-02-27
Remove "viewToken" in favor of the passthrough headers

kellymtrinh:
Reverts airplanedev/js-sdk#300 for `viewToken` which isn't used by anything yet.
I don't need to do this since I can directly use the new headers that were introduced in https://github.com/airplanedev/js-sdk/pull/293/files


packages/airplane/internal/api/client.ts
```diff
@@ -40,19 +40,18 @@ export class Client {
...
     const runID = opts.runID || env?.AIRPLANE_RUN_ID;
-    const viewToken = opts.viewToken || env?.AIRPLANE_VIEW_TOKEN;
-
     // The tokens below are not part of ClientOptions because they are Studio-specific
     // implementation details. The dev server uses the tunnel token to verify requests made to ngrok
     // tunnels, and this should not be exposed to users. This is set by the Studio at runtime.
     const tunnelToken = env?.AIRPLANE_TUNNEL_TOKEN;
     const sandboxToken = env?.AIRPLANE_SANDBOX_TOKEN;
 
-    const headerValue = env?.AIRPLANE_API_HEADERS;
```
kellymtrinh, line 51: @francescaleung I'm adding this since the thing I want to add won't be in process env!

packages/airplane/internal/api/client.ts
```diff
@@ -40,19 +40,18 @@ export class Client {
...
     let headers = {};
-    if (headerValue) {
+    if (opts.headers) {
+      headers = opts.headers;
+    } else if (env?.AIRPLANE_API_HEADERS) {
       try {
-        headers = JSON.parse(headerValue);
+        headers = JSON.parse(env.AIRPLANE_API_HEADERS);
```
francescaleung, line 54: can you make this a union instead of an exclusive or? If there are `AIRPLANE_API_HEADERS` in the environment they should always get passed through regardless of whether or not headers have been explicitly specified.

francescaleung (approved):

PR #307, opened 2023-03-02
Distribute as a hybrid esm/cjs package

leeweisberger:
Distribute the js sdk as a hybrid esm/cjs package.

The library can be imported (esm) or required (cjs).

Previously we were exporting the library as cjs only which limits tree shaking and other features. Consumers should now be able to import the correct module type.

A few other changes described inline...


.github/workflows/tests.yml
```diff
@@ -94,7 +94,7 @@ jobs:
           echo "::endgroup::"
 
           echo "::group::Publishing..."
-          cd "packages/${{matrix.folder}}/dist"
+          cd "packages/${{matrix.folder}}"
```
leeweisberger, line 97: I added `files: ["dist"]` to all of the package.jsons. This should ensure that we only send the `dist` folder, the package.json, readme and license (automatically added). No need to copy these files into the dist folder and then publish the dist folder.

packages/airplane/package.json
```diff
@@ -2,21 +2,18 @@
...
   "description": "JavaScript SDK for writing Airplane tasks",
-  "main": "index.js",
-  "types": "index.d.ts",
   "repository": "airplanedev/js-sdk",
   "author": "hello@airplane.dev",
   "license": "MIT",
   "sideEffects": false,
   "scripts": {
-    "build": "rm -rf dist && mkdir dist && cp ../../LICENSE ../../README.md package.json dist && tsc -p tsconfig.build.json"
+    "build": "rm -rf dist && tsc -p tsconfig.esm.json && tsc -p tsconfig.cjs.json"
```
leeweisberger, line 10: We build two versions of the project into two subfolders

packages/airplane/package.json
```diff
@@ -33,5 +30,123 @@
...
     "ts-dedent": "2.2.0",
     "uuid": "9.0.0"
+  },
+  "files": [
+    "dist"
+  ],
+  "main": "dist/cjs/index.js",
+  "module": "dist/esm/index.js",
+  "types": "dist/cjs/index.d.ts",
+  "exports": {
```
leeweisberger, line 40: This is what modern consumers will use (main/module/types will be ignored and are here for legacy consumers). This tells consumers how to import or require different parts of the code. Only these paths can be required.

Without this, pure-esm consumers cannot import `airplane/api` which is causing us some issues cc @geodn.

packages/airplane/tsconfig.cjs.json
```diff
@@ -0,0 +1,10 @@
+{
+  "extends": "../../tsconfig.base.json",
```
leeweisberger, line 2: All tsconfigs extend from a common, shared base.

packages/workflow-runtime/index.ts
```diff
@@ -0,0 +1 @@
+export * from "./internal";
```
leeweisberger, line 1: I thought this was a bit cleaner rather than trying to pretend the `index.ts` in the internal package is the top level index.ts

geodn (approved):
Should we bump the version as well?

tsconfig.base.json
```diff
@@ -0,0 +1,12 @@
+{
```
colinking, line 1: nice

packages/workflow-runtime/package.json
```diff
@@ -28,9 +27,29 @@
     "@temporalio/interceptors-opentelemetry": "1.4.4",
     "@temporalio/worker": "1.4.4",
     "@temporalio/workflow": "1.4.4",
+    "cross-fetch": "3.1.5",
```
colinking, line 30: The `workflow-runtime` does not use `cross-fetch`. Can you remove this?

packages/workflow-runtime/package.json
```diff
@@ -7,14 +7,13 @@
...
   "scripts": {
-    "build": "rm -rf dist && mkdir -p dist/internal && cp ../../LICENSE ../../README.md package.json dist && tsc"
+    "build": "rm -rf dist && tsc -p tsconfig.esm.json && tsc -p tsconfig.cjs.json"
   },
   "engines": {
     "node": "^14 || ^16 || ^17 || ^18"
   },
   "devDependencies": {
     "@types/node": "18.14.1",
-    "@types/node-fetch": "2.6.2",
```
colinking, line 17: 👍🏻

packages/airplane/package.json
colinking: I'm worried that we'll add new file-based exports and forget to add them here (e.g. we might only test that `airplane.X` works, but not `X` directly).

Do we only need this because you are now shipping the `dist/` folder in the package? We intentionally avoided that so that file-based imports would work. If that behavior does not work with ESM, then this is fine (though is it possible to generate this automatically?).

packages/airplane/internal/runtime/standard.ts
```diff
@@ -1,4 +1,5 @@
 import { HTTPError } from "@airplane/lib";
+import fetch from "cross-fetch";
```
colinking, line 2: Good catch. I expect `AirplaneFile` was not working for Node 14/16 users!

colinking (approved):
Thanks for taking this on! Looks good. My only concern is around the `exports` field and whether we can make that less error-prone.

Consider doing a pre-release first and confirming that this works in both cjs and esm contexts before releasing a patch release. All you have to do is update the version to `0.2.42-1` and CI will automatically release it as `airplane@next`.

packages/airplane/package.json
leeweisberger: This isn't because of the `dist/` folder. This allows these imports to work with pure esm packages like `docs` (where we noticed this issue).

Basically, pure esm requires importing a file **with** the `.js` extension. `import foo` errors. `import foo.js` works.

Currently, if I import `airplane/api`, that errors from an esm project. These exports map `airplane/api` (doesn't end in js) -> `dist/esm/api.js"` (does end in js)

leeweisberger: I'm not sure if there is an easy solution to checking this... I will look some stuff up. I'm going to merge to next to test and unblock george in the meantime

PR #308, opened 2023-03-02
Un-nest cjs output so older consumers can use it

leeweisberger:
In a previous pr, we set up our output to look like this
```
dist/
  esm/
    index.js
  cjs
    index.js
```
And then we added `exports` to the package.json to tell consumers where to import their files from...

Turns out (we're learning) that some consumers don't read the `exports` field. namely, you have to set `moduleResolution` to a new strategy to take advantage of exports. Good read [here](https://stackoverflow.com/questions/71463698/why-we-need-nodenext-typescript-compiler-option-when-we-have-esnext).

What we have to do instead is to keep the cjs exports at the top level for consumers who don't understand cjs (basically the way things were before) and then nest the esm outputs in a directory that newer consumers can reach for:

```
esm/
  index.js
index.js (cjs)
```

This should satisfy older consumers with a structure that looked like it did before, and newer consumers can get the benefits of esm modules.


colinking (approved):
Good catch.

PR #314, opened 2023-03-08
Add support for file uploads

ericliuche:


packages/airplane/internal/api/client.ts
```diff
@@ -290,6 +291,59 @@ export class Client {
...
+      return resp;
+    });
+  }
+
+  async uploadFileContents(
+    url: string,
+    payload: Blob,
+    opts: { abortSignal?: AbortSignal } = {}
+  ): Promise<void> {
+    const controller = new AbortController();
```
ericliuche, line 319: we cant use the fetcher client here cause the url isn't an airplane URL

kellymtrinh: Can you override the fetcher host?

ericliuche: we could, but there's also other issues :/ the fetcher does a ton of airplane specific things like parsing for certain headers and it also always expects a response 
```
      // If we got a 2xx status code, we can return successfully.
      if (response.status >= 200 && response.status < 300) {
        if (isJSONResponse(response)) {
          return (await response.json()) as TOutput;
        }
        return (await response.text()) as TOutput;
      }
```
which in this case we don't

kellymtrinh: Hmm I guess it might also send a lot of unneeded fields too

kellymtrinh (approved):

packages/airplane/package.json
```diff
@@ -27,6 +28,7 @@
     "@opentelemetry/api": "1.2.0",
     "fast-equals": "4.0.3",
     "humanize-string": "3.0.0",
+    "mime": "^3.0.0",
```
kellymtrinh, line 31: Would double check that this doesn't need to be added to lib like how humanize-string did

ericliuche: ill double check, but i think it should be fine cause `mime` isn't esm only

PR #320, opened 2023-03-17
Support getting outputs for runs in self-hosted storage zones

yolken-airplane:
## Description
This change updates the standard runtime in the js-sdk to support getting outputs for runs that are in self-hosted storage zones. For these, the SDK needs to hit the API to get the zone info (see https://github.com/airplanedev/airport/pull/4245) and then use the information in there to hit the user-hosted agent server. 

A customer noticed that this wasn't working and wrote in to complain :-).

Note that this change doesn't fix getting outputs in workflow activities- I'll test that out and follow-up with a fix in a separate change (if needed).

## Testing
Testing completed via running the standard runtime example locally:

Screenshot 2023-03-17 at 1 13 48 PM


packages/airplane/internal/api/client.ts
```diff
@@ -182,8 +182,31 @@ export class Client {
...
     runID: string,
-    opts: { abortSignal?: AbortSignal } = {}
+    opts: { abortSignal?: AbortSignal; useZone?: boolean } = {}
   ): Promise<O> {
+    if (opts.useZone) {
+      // We need to get the run zone, then use that to get the outputs since they're
+      // not stored in the airplane API.
+      const zone = await this.getRunZone(runID, { abortSignal: opts.abortSignal });
+      const { output } = await this.fetcher.get<{ output: O }>(
+        "/v0/dp/runs/getOutputs?",
```
colinking, line 192:
```suggestion
        "/v0/dp/runs/getOutputs",
```

packages/lib/internal/fetcher.ts
```diff
@@ -35,6 +35,8 @@ export type FetcherOptions = {
 export type RequestOptions = {
   // External signal for when to abort the request.
   abortSignal?: AbortSignal;
+  // A custom host override. Intended only for usage with self-hosted storage.
+  hostOverride?: string;
```
colinking, line 39: Similar to `headers` that is available in both here and `FetcherOptions`, no need for the `override` suffix.

```suggestion
  // Optionally override the host used in this request. Takes precedence over the fetcher's host.
  //
  // For example, this can be used to send a request to a self-hosted storage zone instead of the main Airplane API.
  host?: string;
```

packages/airplane/internal/api/client.ts
```diff
@@ -205,9 +228,19 @@ export class Client {
       status: RunStatus;
       paramValues: P;
       taskID?: string;
+      zoneID?: string;
     }>("/v0/runs/get", { id: runID }, { abortSignal: opts.abortSignal });
   }
 
+  async getRunZone(runID: string, opts: { abortSignal?: AbortSignal } = {}) {
```
colinking, line 235:
```suggestion
  async getZoneByRunID(runID: string, opts: { abortSignal?: AbortSignal } = {}) {
```

packages/airplane/internal/api/client.ts
yolken-airplane: Fixed.

packages/airplane/internal/api/get.ts
```diff
@@ -11,7 +11,8 @@ export const getRun = async <
 ): Promise<Run<TParamValues, TOutput>> => {
   const client = new Client(opts);
   const run = await client.getRun<TParamValues>(runID);
-  const runOutput = await client.getRunOutput<TOutput>(runID);
+  const useZone = typeof run.zoneID === "string" && run.zoneID.length > 0;
```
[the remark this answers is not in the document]
yolken-airplane: Ahh, I love javascript, haha. Fixed.

packages/airplane/internal/runtime/standard.ts
```diff
@@ -81,7 +81,8 @@ export const runtime: RuntimeInterface = {
         return null;
       }
 
-      const output = await client.getRunOutput<Output>(runID);
+      const useZone = typeof run.zoneID === "string" && run.zoneID.length > 0;
```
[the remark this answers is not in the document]
yolken-airplane: Fixed.

packages/lib/internal/fetcher.ts
yolken-airplane: Fixed.

packages/airplane/internal/api/client.ts
yolken-airplane: Fixed.

packages/airplane/internal/api/client.ts
```diff
@@ -182,8 +182,31 @@ export class Client {
 
   async getRunOutput<O = unknown>(
     runID: string,
-    opts: { abortSignal?: AbortSignal } = {}
+    opts: { abortSignal?: AbortSignal; useZone?: boolean } = {}
   ): Promise<O> {
+    if (opts.useZone) {
+      // We need to get the run zone, then use that to get the outputs since they're
+      // not stored in the airplane API.
+      const zone = await this.getRunZone(runID, { abortSignal: opts.abortSignal });
```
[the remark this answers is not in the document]
yolken-airplane: Leaving this as a TODO for now.

PR #327, opened 2023-03-22
Try fixing workflow error bubbling

kellymtrinh:
Publish a next version that fixes workflows error bubbling.


leeweisberger (approved):
not really sure why it's -3 instead of -1

kellymtrinh:
> not really sure why it's -3 instead of -1

It's just in case the caching for our team interferes since I used -1 and -2 to deploy a version before before 😭

PR #330, opened 2023-03-23
Fix workflow bubbling up error

kellymtrinh:
The instanceof check still isn't working for workflows.

Trying to avoid bumping so many package versions, so releasing this one as next. I'll bump this after confirming this works.


packages/workflow-runtime/internal/activities.ts
```diff
@@ -296,7 +296,9 @@ export const uploadFileContents: typeof uploadFileContentsActivity = async (url,
 };
 
 const serializeAsTemporalError = (err: unknown) => {
-  if (err instanceof HTTPError) {
+  // instanceof is not a reliable check for HTTPError,
+  // so we check that the error name is HTTPError instead.
+  if (err instanceof Error && err.name === "HTTPError") {
```
kellymtrinh, line 301: `err instanceof Error` This is just to check that it has the fields message, name, etc. but I'm pretty sure the basic instanceof Error works. It doesn't have the same issue as custom errors, and probably wouldn't run into issues if it was because of multiple airplane/lib types

PR #338, opened 2023-03-28
Add convertToDef function on airplane.task and airplane.workflow

ameyakhare:
Resolves CAP-2581

Adds a function to the task object to convert its config to a task definition that lib understands. This will allow us to make changes in the js-sdk that are automatically versioned/compatible with any version of lib.


packages/airplane/internal/config.ts
```diff
@@ -520,3 +531,160 @@ const normalizeConfig = <TParams extends InputParams>(
...
+const convertToDef = <TParams extends InputParams>(
+  config: TaskConfig<TParams> | SQLTaskConfig<TParams>,
+  kind: TaskKind
+): ((fileConfig: { relEntrypoint: string; exportName: string }) => TaskDef<TParams>) => {
+  return (fileConfig: { relEntrypoint: string; exportName: string }) => {
+    const taskDef: TaskDef<TParams> = {
+      slug: config.slug,
+      name: config.name,
+      description: config.description,
+      parameters: config.parameters
```
ericliuche, line 635: use `SerializeParams`

packages/airplane/internal/config.ts
```diff
@@ -520,3 +531,160 @@ const normalizeConfig = <TParams extends InputParams>(
...
+
+/**
+ * The configuration for a SQL Task. To learn more, see the
+ * [Task definition reference](https://docs.airplane.dev/tasks/task-definition).
+ *
+ * Builtin task bindings are currently in beta.
+ */
+export type SQLTaskConfig<TParams extends InputParams> = TaskConfig<TParams> &
+  Omit<TaskConfig<TParams>, "resources" | "runtime"> & {
+    sql: {
```
ericliuche, line 543: move up a level, outside `sql`

packages/airplane/internal/config.ts
```diff
@@ -520,3 +531,160 @@ const normalizeConfig = <TParams extends InputParams>(
...
+    resource: string;
+    query: string;
+    queryArgs?: Record<string, unknown>;
+    transactionMode?: TransactionMode;
+  };
+  // TODO: move this into the node field
+  entrypointFunc?: string;
+};
+
+const convertToDef = <TParams extends InputParams>(
```
ericliuche, line 626: merge the logic in `normalizeConfig` here

packages/airplane/internal/config.ts
```diff
@@ -520,3 +531,160 @@ const normalizeConfig = <TParams extends InputParams>(
...
   };
 };
+
+/**
+ * The configuration for a SQL Task. To learn more, see the
+ * [Task definition reference](https://docs.airplane.dev/tasks/task-definition).
+ *
+ * Builtin task bindings are currently in beta.
+ */
+export type SQLTaskConfig<TParams extends InputParams> = TaskConfig<TParams> &
```
ameyakhare, line 541: @colinking eyes here for bindings and configs for sql tasks

packages/airplane/internal/config.ts
```diff
@@ -520,3 +531,160 @@ const normalizeConfig = <TParams extends InputParams>(
...
+        kind === "node"
+          ? {
+              entrypoint: fileConfig.relEntrypoint,
+              envVars: config.envVars,
+            }
+          : undefined,
+      sql:
+        kind === "sql"
+          ? {
+              resource: sqlConfig.resource,
```
ameyakhare, line 679: @colinking typecasted const var access since the config object passed in could be a TaskConfig or SQLTaskConfig

packages/airplane/internal/params/params_serialized.ts
```diff
@@ -105,14 +118,13 @@ export const serializeParams = (params: InputParams): SerializedParams => {
     const optional = param.required === false ? true : false;
     serializedParams.push({
       slug: slug,
-      // TODO: add a slug -> name conversion, e.g. my_param -> My Param
-      name: param.name ? param.name : slug,
+      name: param.name ? param.name : humanizeString(slug),
```
colinking, line 121: We already do this in `normalizeConfig`. Why do it here, too?

packages/airplane/internal/config.ts
```diff
@@ -520,3 +530,146 @@ const normalizeConfig = <TParams extends InputParams>(
     name: config.name ? config.name : humanizeString(config.slug),
   };
 };
+
+/**
+ * The configuration for a SQL Task. To learn more, see the
+ * [Task definition reference](https://docs.airplane.dev/tasks/task-definition).
+ *
+ * Builtin task bindings are currently in beta.
```
colinking, line 538:
```suggestion
 * Declaring SQL tasks with JavaScript is currently in beta.
```

Unrelated, but this made me notice `Airplane Views are currently in beta.` is outdated -- can you remove that?

packages/airplane/internal/config.ts
```diff
@@ -520,3 +530,146 @@ const normalizeConfig = <TParams extends InputParams>(
...
+ */
+export type SQLTaskConfig<TParams extends InputParams> = TaskConfig<TParams> &
+  Omit<TaskConfig<TParams>, "resources" | "runtime"> & {
+    resource: string;
+    query: string;
+    queryArgs?: Record<string, unknown>;
+    transactionMode?: TransactionMode;
```
colinking, line 545: Add documentation (with examples and docs links) for these fields. See the types above for examples, e.g. `TaskConfig`.

packages/airplane/internal/config.test.ts
```diff
@@ -320,4 +320,49 @@ describe("config", () => {
...
+      },
+      () => {
+        return "Hello!";
+      }
+    );
+
+    const def = (t as any).__airplane.convertToDef({ relEntrypoint: "foo", exportName: "bar" });
+    expect(def.slug).toEqual("hello");
+    expect(def.name).toEqual("Hello");
+    expect(def.node.entrypoint).toEqual("foo");
```
colinking, line 347: This should not be set.

packages/airplane/internal/config.test.ts
```diff
@@ -320,4 +320,49 @@ describe("config", () => {
 
     expect((vv as any).__airplane.config.name).toBe(`Name!`);
   });
+
+  test("config serialized to definition", () => {
```
colinking, line 324: Nice. Add a view and SQL test, too.

packages/airplane/internal/config.ts
```diff
@@ -520,3 +530,146 @@ const normalizeConfig = <TParams extends InputParams>(
...
+  wrappedF.__airplane = {
+    type: "task",
+    config: normalizeConfig(config),
+    convertToDef: convertToDef(config, "sql"),
+  };
+  return wrappedF;
+};
+
+// Attach sql task binding on task object.
+task.sql = sqlTask;
```
colinking, line 575: Add type tests for `airplane.task.sql` in `config.test-d.ts`. This ensures that we're producing readable (and correct) types. These tests are checked via `eslint-plugin-expect-type`. `eslint` will automatically fix any of the `// ^?` comments to include the corresponding type.

colinking (commented):
First pass.

packages/airplane/internal/config.test.ts
colinking: Ah, I see this is a parameter. Nevermind.

packages/airplane/internal/config.ts
```diff
@@ -520,3 +530,146 @@ const normalizeConfig = <TParams extends InputParams>(
...
+  type: string;
+  name?: string;
+  description?: string;
+  required?: boolean;
+  default?: InputParamValue;
+  regex?: string;
+  options?: (InputParamValue | SelectOption<InputParamValue>)[];
+};
+
+export const serializedParamToDefParam = (param: SerializedParam): ParamDef => {
```
colinking, line 590: unexport

packages/airplane/internal/config.ts
```diff
@@ -520,3 +530,146 @@ const normalizeConfig = <TParams extends InputParams>(
...
 };
+
+/**
+ * The configuration for a SQL Task. To learn more, see the
+ * [Task definition reference](https://docs.airplane.dev/tasks/task-definition).
+ *
+ * Builtin task bindings are currently in beta.
+ */
+export type SQLTaskConfig<TParams extends InputParams> = TaskConfig<TParams> &
+  Omit<TaskConfig<TParams>, "resources" | "runtime"> & {
```
colinking, line 541: We don't support `envVars` for SQL tasks, either.

packages/airplane/internal/config.ts
```diff
@@ -520,3 +530,146 @@ const normalizeConfig = <TParams extends InputParams>(
...
   };
 };
+
+/**
+ * The configuration for a SQL Task. To learn more, see the
+ * [Task definition reference](https://docs.airplane.dev/tasks/task-definition).
+ *
+ * Builtin task bindings are currently in beta.
+ */
+export type SQLTaskConfig<TParams extends InputParams> = TaskConfig<TParams> &
```
colinking, line 540: This is bringing back the omitted fields, e.g. `resources`:

CleanShot 2023-04-05 at 15 50 32

```suggestion
export type SQLTaskConfig<TParams extends InputParams> =
```

packages/airplane/internal/config.ts
```diff
@@ -520,3 +530,146 @@ const normalizeConfig = <TParams extends InputParams>(
...
+  const wrappedF = async (
+    paramValues?: Record<string, unknown>,
+    executeOptions?: ClientOptions
+  ) => {
+    const run = await execute<Awaited<{ Q1: QType<TRow> }>>(
+      config.slug,
+      paramValues as InputParamValues<DefaultEmpty<TParams>>,
+      executeOptions
+    );
+    return run as Run<
```
colinking, line 560: This should be `BuiltInRun` (which does not have a `taskID` field). You'll also need to add a `WrappedBuiltInRunFunc` so that this task returns a `BuiltInRun`.

packages/airplane/internal/config.ts
```diff
@@ -520,3 +530,146 @@ const normalizeConfig = <TParams extends InputParams>(
...
+    name: param.name,
+    description: param.desc,
+    required: !param.constraints.optional ? true : undefined,
+    default: param.default,
+    regex: param.constraints.regex,
+    options: param.constraints.options,
+  };
+};
+
+type TaskDef<TParams extends InputParams> = {
```
colinking, line 603: Do you need this type? IIUC, this is only used in one place (`convertToDef`).

packages/airplane/internal/config.ts
colinking: +1. Can you just call `normalizeConfig` from `convertToDef` instead of duplicating the dedent/humanize logic?

colinking (changes requested):
Generally looks good. I just want to approve the type test changes.

If you haven't already, let's give this a good test in studio/builds before we start using this `convertToDef` method.

packages/airplane/internal/config.ts
```diff
@@ -342,6 +349,7 @@ export const task = <TParams extends InputParams, TOutput>(
       sdkVersion: VERSION,
     },
     baseFunc,
+    convertToDef: convertToDef(config, "node"),
```
colinking, line 352: It seems confusing to have both `config` and `convertToDef`. It looks like Views only uses `config.slug`. Discovery is presumably the only other place looking at `config` which would get migrated to `convertToDef`?

packages/airplane/internal/config.ts
```diff
@@ -528,3 +537,96 @@ const normalizeConfig = <TParams extends InputParams>(
...
+
+  node?: {
+    entrypoint: string;
+    envVars?: Record<string, string | { config: string } | { value: string }>;
+  };
+  // TODO: move this into the node field
+  entrypointFunc?: string;
+};
+
+const convertToDef = <TParams extends InputParams>(
```
colinking, line 594: Does this need to be a function? What if it was `__airplane.def` but without `node.entrypoint` or `entrypointFunc` set, and it was up to discovery to set those fields? That would allow Views to switch from `config.slug` to `def.slug` without having to call this function with fake values just to get the slug.

packages/airplane/internal/config.ts
colinking: Could we remove `config` in this PR? That way we only have one of these. This is assuming we update Views and discovery first with logic for using `convertToDef` if set, else `config`.

packages/airplane/internal/config.ts
```diff
@@ -528,3 +537,96 @@ const normalizeConfig = <TParams extends InputParams>(
...
+    const taskDef: TaskDef<TParams> = {
+      slug: config.slug,
+      name: config.name,
+      description: config.description ? dedent(config.description) : undefined,
+      parameters: config.parameters
+        ? serializeParams(config.parameters).map(serializedParamToDefParam)
+        : undefined,
+      runtime: config.runtime === "workflow" ? "workflow" : config.runtime ? "" : undefined,
+      resources: config.resources,
+      timeout: config.timeout,
```
colinking, line 610: Maybe just spread out config with `...config`? Same idea as how it works for `__airplane.config` above. That way we don't need to add new task fields here unless their JS format differs from their definition format.

You can do the same for the type, ala `type TaskDef<TParams extends InputParams> = Omit<TaskConfig<TParams>, "parameters"> & { parameters: ParamDef[] }`

packages/airplane/internal/config.ts
```diff
@@ -528,3 +537,96 @@ const normalizeConfig = <TParams extends InputParams>(
...
+  parameters?: ParamDef[];
+  runtime?: "" | "workflow";
+  resources?: Record<string, string> | string[];
+  configs?: string[];
+  timeout?: number;
+  constraints?: Record<string, string>;
+  requireRequests?: boolean;
+  allowSelfApprovals?: boolean;
+  restrictCallers?: string[];
+  schedules?: Record<string, Schedule<InputParamValues<TParams>>>;
```
colinking, line 580: You'll need to serialize schedule parameter values, too. See how `serializeParams` calls `serializeParamValue` on `default`.

packages/airplane/internal/params/params_serialized.ts
```diff
@@ -91,6 +91,18 @@ const serializeParamType = (
   }
 };
 
+export const serializeParamDef = (
```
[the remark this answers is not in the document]
colinking: Ah, nm. The goal of `params_serialized.ts` is to represent the API format for parameters. However, I suppose there's a separate _definition_ format.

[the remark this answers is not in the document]
colinking: Let's keep the definition format out of this file, then, since it's unrelated to the serialized API format. How about we just inline this in `serializedParamToDefParam`?

packages/airplane/internal/config.ts
```diff
@@ -528,3 +537,96 @@ const normalizeConfig = <TParams extends InputParams>(
     name: config.name ? config.name : humanizeString(config.slug),
   };
 };
+
+type TaskKind = "node" | "view";
+
+type ParamDef = {
```
colinking, line 543: To align with the other parameter types, do `DefParam` (e.g. InputParam/SerializedParam/...)

packages/airplane/internal/config.ts
```diff
@@ -528,3 +537,96 @@ const normalizeConfig = <TParams extends InputParams>(
...
+  type: string;
+  name?: string;
+  description?: string;
+  required?: boolean;
+  default?: InputParamValue;
+  regex?: string;
+  options?: (InputParamValue | SelectOption<InputParamValue>)[];
+};
+
+const serializedParamToDefParam = (param: SerializedParam): ParamDef => {
```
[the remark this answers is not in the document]
colinking: nm see comment above

PR #349, opened 2023-04-14
Fix file upload for workflows

ericliuche:
To pass the payload into the workflow activity, it has to be serializable. To do so, we first convert the blob into a UInt8Array and then convert it back when we go to upload


kellymtrinh (approved):
Wow nice find!

PR #350, opened 2023-04-14
Resuse a shared tcp connection to reduce overhead

ericliuche:
Before, when calling many many long running tasks in parallel, we would see that tasks would fail due to a client timeout. `Request timed out after 10000ms.`. When investigating the API requests in our backend, none actually took longer than 10s. The theory here is that we're exhausting the TCP connection pool and the delay waiting for a new TCP connection is what is exhausting our timeout.

To address this, we use a shared http client to reduce overhead with initializing TCP connections with our API. 

This task that used to fail with a request timeout (https://web.airstage.app/runs/run20230414zuwhev8dds1?__env=prod), now passes: (https://web.airstage.app/runs/run20230414zzjxa9z6v57?__env=prod). 

```
export default airplane.task(
  {
    slug: "test_error",
    name: "test error",
    parameters: {
      times: "integer",
      duration: "shorttext",
    },
  },
  // This is your task's entrypoint. When your task is executed, this
  // function will be called.
  async (params) => {
    const children = new Array(params.times).fill(1);
    const output = await Promise.all(
      children.map(async (c) => {
        const child = await airplane.execute("test_sleep", { duration: params.duration });
        return child.output;
      })
    );
    return output.flat();
  }
);

export const sleepTest = airplane.task(
  {
    slug: "test_sleep",
    name: "test_sleep",
    parameters: {
      duration: "shorttext",
    },
  },
  // This is your task's entrypoint. When your task is executed, this
  // function will be called.
  async (params) => {
    await airplane.sleep(params.duration);
    return "wat";
  }
);
```

I also deployed a view and workflow and tested that the fetcher changes succeeded for them. also tested in studio


packages/lib/internal/fetcher.ts
```diff
@@ -69,6 +74,11 @@ export class Fetcher {
     }
 
     this.opts = opts;
+    if (opts.host.startsWith("http://")) {
```
kellymtrinh, line 77: Are these the same?

kellymtrinh: Nvm lol missed the http/https part

kellymtrinh (approved):

packages/lib/internal/fetcher.ts
```diff
@@ -1,3 +1,6 @@
+import http from "http";
+import https from "https";
```
colinking, line 2: Import just the `Agent` from both packages. That should reduce bundle size by not importing the entire `http/https` package.

```suggestion
import { Agent as httpsAgent } from "https";
```

packages/lib/internal/fetcher.ts
```diff
@@ -69,6 +74,11 @@ export class Fetcher {
     }
 
     this.opts = opts;
+    if (opts.host.startsWith("http://")) {
+      this.agent = new http.Agent({ keepAlive: true });
+    } else {
+      this.agent = new https.Agent({ keepAlive: true });
```
colinking, line 80: Won't this fail in a browser context since the `https/http` packages aren't defined?

Looking at similar implementations, they only use an agent in Node: https://github.com/stainless-sdks/airplane-node/blob/ae1f598ebda4467823aac574373cfb632a905a63/core.ts#L17-L28

packages/lib/internal/fetcher.ts
```diff
@@ -282,6 +294,12 @@ const fetchWithRetries = async <TOutput>(req: FetchWithRetriesRequest): Promise<
         signal: controller.signal as any,
         mode: req.mode,
         credentials: req.credentials,
+        // We pass in a custom agent so that we can try and use the same connection for multiple
+        // requests. The agent option is only available in node-fetch so it is not exposed
+        // in cross-fetch's interface. Despite this, when cross-fetch pollyfills fetch with node-fetch,
+        // passing in `agent` will work.
+        // @ts-ignore
```
colinking, line 301:
```suggestion
        // We pass in a custom agent to re-use connections across multiple requests. This is only
        // supported in Node.js.
        // @ts-ignore
```

colinking (approved):
Make sure to test this in Views first! Good job figuring this out.

colinking:
> The theory here is that the event loop is kept busy performing too many operations that when we go to communicate with our API, the timeout has already expired.

Also to clarify, the event loop should not be the issue (that should be able to handle thousands of operations per second). Presumably we're exhausting the TCP connection pool and the delay waiting for a new TCP connection is what is exhausting our timeout.

packages/lib/internal/fetcher.ts
ericliuche: @colinking could I get access to the repo? I don't think I have permissions for it

PR #354, opened 2023-04-25
Add permissions to inline task definition

rose-huang:
Allow users to define permissions for tasks


packages/airplane/internal/config.test-d.ts
```diff
@@ -594,6 +594,46 @@ const _task = async () => {
...
     );
   }
+
+  {
+    // Task with explicit permissions.
+    const _myTask1 = task(
+      {
+        slug: "greet",
+        parameters: {
+          shorttext: "shorttext",
```
leeweisberger, line 604: Nice! Let's get rid of parameters b/c we aren't testing for that in this test

leeweisberger (approved):
Cool. This I think we should hold off on merging until we are ready to launch this

PR #358, opened 2023-04-27
Add defaultRunPermissions to js sdk

leeweisberger:
Add defaultRunPermissions to the js-sdk so it can be defined inline for a JS task.

This will not be merged until we're ready to release run permissions.


packages/airplane/internal/config.ts
```diff
@@ -140,6 +141,11 @@ export type TaskConfig<TParams extends InputParams> = {
    * @default []
    */
   restrictCallers?: TaskCaller[];
+  /**
+   * Manage who has permissions for new runs of this task.
```
rose-huang, line 145: maybe mention that this can be changed on the specific runs page?

rose-huang (approved):

packages/airplane/internal/config.ts
```diff
@@ -69,6 +69,7 @@ export type Schedule<TParamValues extends InputParamValues> = {
     });
 
 export type TaskCaller = "task" | "view";
+export type DefaultRunPermission = "task-viewers" | "task-participants";
```
colinking, line 72: Since this is task-specific, do we need the `task-` prefix?

packages/airplane/internal/config.ts
colinking, line 145: Could you copy over the descriptions from the? I wasn't sure what "task-participants" was without consulting the website.

Also, add a docs link.

CleanShot 2023-05-01 at 11 43 28

PR #361, opened 2023-05-04
Add SDK version to task config

ericliuche:


scripts/version.sh
```diff
@@ -13,6 +13,7 @@ yarn workspace @airplane/lib version --no-git-tag-version --new-version $NEW_VER
 # warning "airplane" is already in "dependencies". Please remove existing entry first before adding it to "peerDependencies".
 # yarn workspace @airplane/workflow-runtime add --peer airplane@$NEW_VERSION
 sed -i'.bak' -e "s/\"airplane\": \"$OLD_VERSION\"/\"airplane\": \"$NEW_VERSION\"/" ./packages/workflow-runtime/package.json
+sed -i -e "s/VERSION = \"$OLD_VERSION\"/VERSION = \"$NEW_VERSION\"/" ./packages/airplane/internal/version.ts
```
leeweisberger, line 16: Are we relying on this being called? There's no automated process that calls this. You need to add a gh action precommit step that does this

leeweisberger (changes requested):

leeweisberger (approved):
We should really have an additional check that ensures that this value is correct. I don't have to call bumpversion

PR #366, opened 2023-05-17
Bubble up network errors

calufornia:
When the Studio dev server is not running, requests made to the studio host error with `TypeError: Network request failed
    at xhr.onerror (browser-ponyfill.js:480:1)`. Unfortunately, it does not seem like we can differentiate between a host not being up vs. flakiness in a user's connection (the error does not contain any more info). Therefore, we add an option to the fetcher to let the caller decide whether to retry network errors (in the case of Studio, it's not likely that the error is due to flakiness and so we'll set it to false). This PR makes it so that in the case that this configuration is set to false, the request will fail right away instead of waiting for the max number of attempts to be hit.


calufornia:
@colinking LMK if you know how to test this—I tried using nock's `retryWithError` modifier but that seems to always wrap the argument in a `FetchError`

colinking (changes requested):
Heads up that the `fetch` implementation will vary between Node (where tests run) and the browser. 

There's a bit of info on `node-fetch` [error handling here](https://github.com/node-fetch/node-fetch/blob/main/docs/ERROR-HANDLING.md). Sounds like you might get a `FetchError` back. Can you add a jest test that tests the behavior of connecting to a port that is not listening?

As for the browser, you'll just want to import this into `web` and try it. It should be the same approach as testing JS SDK with Views: https://www.notion.so/airplanedev/Testing-JS-SDK-in-Views-8ecca48891e0463e9b3c1999ff1631fe

calufornia:
> Heads up that the `fetch` implementation will vary between Node (where tests run) and the browser.
> 
> There's a bit of info on `node-fetch` [error handling here](https://github.com/node-fetch/node-fetch/blob/main/docs/ERROR-HANDLING.md). Sounds like you might get a `FetchError` back. Can you add a jest test that tests the behavior of connecting to a port that is not listening?
> 
> As for the browser, you'll just want to import this into `web` and try it. It should be the same approach as testing JS SDK with Views: https://www.notion.so/airplanedev/Testing-JS-SDK-in-Views-8ecca48891e0463e9b3c1999ff1631fe

Ah yeah meant in a unit test (I've tested this in web)—will try adding a test that covers that scenario

calufornia:
@colinking Added a jest test for this.

> Based on [this SO comment](https://stackoverflow.com/a/61113372), the behavior may also differ when its due to a CORS error (this is easy to test by hitting a path the CLI is not listening on as the `OPTIONS` will 404 causing a CORS error). Also give this a try in Safari, too, to see what we get there.

Works in Safari, but the CORS error appears to be the same 🙁

PR #371, opened 2023-05-25
Drop Node 14 support

colinking:
Drops Node 14 support. For folks still using Node 14, they should use `0.2.60`.

Node 14 support is blocking us from upgrading deps that disallow Node 14.

I'm planning to release this as a patch release since Node 14 is EOL-ed, so this is not a breaking change for anyone unless they are an EOL-ed Node version. sgty @ericliuche?


colinking:
* main
  * **PR #371** <a href="https://app.graphite.dev/github/pr/airplanedev/node-sdk/371" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/airplanedev/node-sdk/371?utm_source=stack-comment).

packages/lib/internal/fetcher.ts
```diff
@@ -302,7 +302,6 @@ const fetchWithRetries = async <TOutput>(req: FetchWithRetriesRequest): Promise<
         method: req.method,
         headers: req.headers,
         body: req.body,
-        // TODO: remove any cast once we drop support for node 14
         // eslint-disable-next-line @typescript-eslint/no-explicit-any
         signal: controller.signal as any,
```
colinking, line 306: Not sure what this referred to.

ameyakhare: probably this line?: signal: controller.signal as any,

ameyakhare (approved):

packages/lib/internal/fetcher.ts
colinking: Yeah, but removing the `as any` still produces a type error as the two `AbortSignal` types don't line up.

ericliuche (approved):
sgtm!

PR #388, opened 2023-06-20
Support list params with 'multi' parameter option

ameyakhare:
Resolves CAP-2815

Adds support for list parameters with `multi` field. Parameters with `multi: true` are resolved to a list of the underlying function input param type, while all others remain unchanged.

This works now, with correct typings:

```
export const listTest = airplane.task(
  {
    slug: "list_test",
    name: "List Test",
    parameters: {
      list_param: {
        type: "shorttext",
        multi: true,
      }
    }
  },
  async (params) => {
    console.log(params.list_param)

    const { list_param } = await airplane.prompt({
      list_param: {
        type: "shorttext",
        multi: true,
        options: [
          "option1",
          "option2",
        ]
      },
    })
    console.log(list_param)
  }
)
```


colinking (changes requested):

packages/airplane/internal/config.test-d.ts
```diff
@@ -333,6 +333,135 @@ const _task = async () => {
     //       }
   }
 
+  // Task with each parameter: json.
```
colinking, line 336: I pushed up a commit with a few more test cases! For context, I just copied what was there and switched the param types, ran `yarn lint --fix` to get it to update the type comments, then reviewed the type comments to make sure they were what I expected.

packages/airplane/internal/params/params_output.test.ts
```diff
@@ -24,7 +32,14 @@ describe("deserializeParamValues", () => {
         f: 1.1,
         g: "2019-08-05",
         h: "2018-08-08T00:00:00-04:00",
-        i: "extraneous param that should be included",
+        i: { __airplaneType: "configvar", name: "name", value: "value" },
+        j: { __airplaneType: "upload", id: "id", url: "url" },
+        k: { foo: "bar" },
+        l: [
+          { __airplaneType: "upload", id: "id", url: "url" },
+          { __airplaneType: "upload", id: "id2", url: "url2" },
```
colinking, line 40: nice

packages/airplane/internal/params/params_serialized.ts
```diff
@@ -73,15 +73,29 @@ export const serializeParamValues = (paramValues: InputParamValues): SerializedP
   return Object.fromEntries(newParamEntries);
 };
 
-export const serializeParamValue = (paramValue: InputParamValue): SerializedParamValue => {
+export const serializeParamValue = (
+  paramValue: InputParamValue | InputParamValue[],
+  multi?: boolean
+): SerializedParamValue | SerializedParamValue[] => {
+  if (multi) {
+    if (paramValue === undefined || paramValue === null) {
```
colinking, line 81:
```suggestion
    if (paramValue == null) {
```

packages/airplane/internal/params/params_serialized.ts
```diff
@@ -73,15 +73,29 @@ export const serializeParamValues = (paramValues: InputParamValues): SerializedP
...
+      return paramValue;
+    }
+    return ((paramValue as InputParamValue[]) ?? []).map((v) => {
```
colinking, line 84: You shouldn't need this since you are already checking if it's `null`

```suggestion
    return (paramValue as InputParamValue[]).map((v) => {
```

packages/airplane/internal/params/params_serialized.ts
```diff
@@ -73,15 +73,29 @@ export const serializeParamValues = (paramValues: InputParamValues): SerializedP
...
+      return serializeParamValueSingle(v);
+    });
```
colinking, line 86:
```suggestion
    return ((paramValue as InputParamValue[]) ?? []).map(serializeParamValueSingle);
```

packages/airplane/internal/config.test-d.ts
colinking: Can you add tests for json/multi to `prompt.test-d.ts`, too? DM me if you have troubles with it.

colinking (approved):
Looks great!

PR #391, opened 2023-06-26
Keep param values that don't exist in param def when deserializing

geodn:


packages/airplane/internal/params/params_output.ts
```diff
@@ -19,11 +19,14 @@ export type OutputParamValueByType = Omit<
...
+export const deserializeParamValues = <
+  TParams extends InputParams,
+  TParamValues extends InputToSerializedParamValues<TParams>
+>(
   params: TParams | undefined,
-  paramValues: InputToSerializedParamValues<TParams>
+  paramValues: TParamValues
 ): Resolve<InputToOutputParamValues<TParams>> => {
-  const output = {} as Record<string, unknown>;
+  const output = paramValues as Record<string, unknown>;
```
geodn, line 29: This is the only functional change, everything else is typing

kellymtrinh (approved):

PR #400, opened 2023-07-18
Use messageOption when calling Slack built-in

stepzhou:
This PR starts to use the new `messageOption` param for the plaintext Slack text. It's cleaner to just have all messages go through `messageOption`. I'll add support for all the new options in a follow-up PR.

Shouldn't merge until https://github.com/airplanedev/airport/pull/5168 is deployed

Resolves CAP-2924


packages/airplane/internal/builtins/slack.ts
```diff
@@ -13,7 +13,13 @@ export const message = async (
   const { client } = opts;
   return executeWithResources(
     "airplane:slack_message",
-    { channelName, message: dedent(message) },
+    {
+      channelName,
+      message: "",
```
kellymtrinh, line 18: Can you exclude message instead? or does it need to have it here but as empty?

This should be backwards compatible in the backend right?

kellymtrinh (approved):

packages/airplane/internal/builtins/slack.ts
stepzhou: I didn't make `message` optional in airport so just keeping it as empty string for now. Will remove eventually

PR #401, opened 2023-07-20
Remove AbortController polyfill

colinking:
This PR removes the `AbortController` and `AbortSignal` polyfill we were using. We no longer support Node 14, so we no longer need to polyfill it.

After upgrading `cross-fetch`, we started noticing issues related to https://github.com/Azure/azure-sdk-for-js/issues/5531

## Test plan

Ran `web` with the updated `@airplane/lib` and confirmed it rendered. Deployed the standard and workflow examples and confirmed both executed. Ran a View locally and confirmed it rendered.


colinking:
* main
  * **PR #401** <a href="https://app.graphite.dev/github/pr/airplanedev/node-sdk/401" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/airplanedev/node-sdk/401?utm_source=stack-comment).

packages/lib/internal/fetcher.ts
calufornia, line 1: Remove console.log statements

calufornia (approved):
🔥

packages/lib/internal/fetcher.ts
colinking: bless

PR #409, opened 2023-07-26
Pass messageOption through Slack built-in

stepzhou:
This PR allows user to pass certain Slack web API options to our Slack built-in.

Depends on https://github.com/airplanedev/web/pull/2974

Resolves CAP-2924


packages/airplane/internal/builtins/slack.ts
```diff
@@ -1,23 +1,52 @@
...
-  return executeWithResources(
-    "airplane:slack_message",
-    {
-      channelName,
-      message: dedent(message),
-    },
-    { slack: "res00000000zteamslack" },
-    client,
-  );
+  if (typeof message === "string") {
```
kellymtrinh, line 27: Styling thing: could you conditionally build just the request? That way if the name `airplane:slack_message` or the slack ID `res00000000zteamslack`  ever changes we'll only need to update one spot.

kellymtrinh (approved):

PR #423, opened 2023-08-18
Add flow for conditionally using self-hosted inputs and/or uploads

ameyakhare:
Resolves CAP-3020
Resolves CAP-3022

Adds the flow to create uploads and/or run inputs via self-hosted storage.


packages/airplane/internal/api/client.ts
```diff
@@ -112,6 +124,94 @@ export class Client {
...
+        if (e instanceof HTTPError) {
+          if (e.statusCode >= 400 && e.statusCode <= 500) {
+            e.message = `Failed to execute task "${slug}": ${e.message}`;
+          }
+        }
+        throw e;
+      }
+    });
+
+    // TODO: this substitution conditional is temporary
```
yolken-airplane, line 172: Ok, this is finally fixed in https://github.com/airplanedev/airport/pull/5420. Can you update the flow here to follow that? Basically, you pass both the parameters and param values to the agent server, and that returns a the updated values that can be sent to our API's task execute handler.

yolken-airplane: If the task parameters aren't readily available, I can update the airplane API to return them from `/v0/inputs/pickZone`?

yolken-airplane: See https://github.com/airplanedev/web/pull/3124/files#diff-a4dbe9e894f4d900a01791873e6025ae0ce6fa6f7d78eaec4396d72cc9b81211R97 for the flow in the FE.

yolken-airplane (approved):

PR #435, opened 2023-09-19
Fix zone picking logic for sub-tasks

yolken-airplane:
## Description
This change fixes the `pickZone` logic to use the passed-in slug first, then fall back to the task revision ID (if defined). This is needed to properly handle zones for sub-tasks, and it also matches the logic in the python-sdk (see https://github.com/airplanedev/python-sdk/blob/main/airplane/api/client.py#L427). 

## Testing
Testing completed successfully via deploying and running an example.


packages/airplane/internal/api/client.ts
```diff
@@ -681,9 +681,11 @@ export class Client {
   }
 
   private async pickZone(slug?: string): Promise<PickZoneResponse> {
-    const args = globalThis.process?.env?.AIRPLANE_TASK_REVISION_ID
+    const args = slug
+      ? { taskSlug: slug }
+      : globalThis.process?.env?.AIRPLANE_TASK_REVISION_ID
       ? { taskRevisionID: globalThis.process?.env?.AIRPLANE_TASK_REVISION_ID }
-      : { taskSlug: slug };
+      : {};
```
leeweisberger, line 688: We definitely want an empty object here than just `{ taskRevisionID: globalThis.process?.env?.AIRPLANE_TASK_REVISION_ID }` which would evaluate to `{ taskRevisionID: undefined }`?

leeweisberger (approved):

PR #440, opened 2023-09-25
Support runnerConfigs for js inline task definitions

yolken-airplane:
## Description
This change updates the js-sdk to support setting the `runnerConfig` task property that was added in https://github.com/airplanedev/airport/pull/5624.

## Testing
Testing completed successfully via deploying an example task with `runnerConfig` set and verifying that the task revision was updated as expected in the API.


examples/standard/package.json
```diff
@@ -5,7 +5,7 @@
     "@types/node": "18.17.15"
   },
   "dependencies": {
-    "airplane": "0.2.73"
+    "airplane": "0.2.75"
```
yolken-airplane, line 8: Noticed that some of these subpackge versions were out-of-date, so also updating them in this change.

packages/airplane/internal/config.ts
```diff
@@ -103,6 +103,20 @@ export type PermissionAssignees = {
...
 export type TaskCaller = "task" | "view";
 export type DefaultRunPermission = "task-viewers" | "task-participants";
+
+export type RunnerConfig = {
+  k8sPodSpecPatch?: K8SPodSpecPatch;
+};
+
+export type K8SPodSpecPatch = {
+  /**
+   * Strategic merge patch to apply to Kubernetes runner pod specs. Only applies
```
leeweisberger, line 113: Is there a k8s doc we can link to?

leeweisberger (approved):

packages/airplane/internal/config.ts
yolken-airplane: I'm going to add a link to our docs once those are ready! That will be better than the kube docs.

PR #451, opened 2023-10-16
Support self-hosted displays in js-sdk

yolken-airplane:
## Description
This change updates the js-sdk to support creating self-hosted displays. The flow is similar to inputs- first, we save the display in the agent server, then we save a placeholder in the Airplane API.

## Testing
Testing completed successfully via deploying example locally (in conjunction with https://github.com/airplanedev/airport/pull/5753 and https://github.com/airplanedev/web/pull/3443).


packages/airplane/internal/api/client.ts
```diff
@@ -76,6 +81,17 @@ export class Client {
...
+    this.agentStorageInternalHost =
+      opts.agentStorageInternalHost || env?.AP_AGENT_STORAGE_INTERNAL_HOST;
+
+    // Use self-hosted outputs if both a zone slug and storage internal host are set.
+    this.useSelfHostedOutputs =
+      (opts.useSelfHostedOutputs || !!env?.AP_AGENT_STORAGE_ZONE_SLUG) &&
+      !!this.agentStorageInternalHost;
+
+    // Save the run ID since this is needed for some agent-storage-related calls.
+    this.agentStorageRunID = runID;
```
ameyakhare, line 93: why is a run stored on the client itself? cant we always read this from the env? (assuming this always executes in the context of a run)

ameyakhare (approved):

PR #452, opened 2023-10-18
Support self-hosted prompts in js sdk

yolken-airplane:
## Description
This change updates the js sdk to support self-hosted prompts.

## Testing
Testing completed successfully via deploying locally and verifying that prompts worked correctly in conjunction with https://github.com/airplanedev/web/pull/3458 and https://github.com/airplanedev/airport/pull/5771.


packages/airplane/internal/api/client.ts
```diff
@@ -506,6 +517,65 @@ export class Client {
...
 
+  async createPromptSelfHostedOutputs(
+    prompt: CreatePromptRequest,
+    opts: { abortSignal?: AbortSignal; idempotencyKey?: string } = {},
+  ): Promise<string> {
+    const agentHost = this.agentStorageInternalHost;
+    const runID = this.agentStorageRunID;
+
+    const agentPath = "/v0/dp/prompts/create";
+    const apiPath = "/v0/prompts/create";
```
ameyakhare, line 528: why these vars? afaict they're only used in 1 place

packages/airplane/internal/api/client.ts
```diff
@@ -733,10 +803,37 @@ export class Client {
...
+    // Get the prompt from the API
+    const apiResp = await this.fetcher.get<{
       prompt: Prompt;
     }>("/v0/prompts/get", { id }, { abortSignal: opts.abortSignal });
-    return resp.prompt;
+
+    if (apiResp.prompt.zoneToken) {
+      // There's a zone token set, so use that to get the real schema and values
+      const agentHost = this.agentStorageInternalHost;
+      const runID = this.agentStorageRunID;
```
ameyakhare, line 814: ditto

ameyakhare (approved):

packages/airplane/internal/api/client.ts
yolken-airplane: No idea, I was just copying this pattern from elsewhere in the SDK. :-)

packages/airplane/internal/api/client.ts
yolken-airplane: Removed these.

packages/airplane/internal/api/client.ts
yolken-airplane: Removed the `agentHost` and `runID` vars. The path ones are used in a few places, so keeping them for now.

PR #455, opened 2023-10-20
Stop testing node 16

yolken-airplane:
## Description
This change updates js-sdk to stop testing node 16. It looks like this version is already past its EOL date, and some of our dependencies have started removing support, which is breaking changes like https://github.com/airplanedev/js-sdk/pull/454.

## Testing
Testing not required because this is just removing a test environment.


colinking (changes requested):
You'll also need to remove the other references to Node 16, e.g. in `engines`: https://github.com/search?q=repo%3Aairplanedev%2Fjs-sdk%2016&type=code

Can you also update the PR title to "Remove Node 16 support"?

yolken-airplane:
> You'll also need to remove the other references to Node 16, e.g. in `engines`: https://github.com/search?q=repo%3Aairplanedev%2Fjs-sdk%2016&type=code
> 
> Can you also update the PR title to "Remove Node 16 support"?

Ahh, ok, will do. Thanks!

colinking (approved):

PR #463, opened 2023-11-13
Support allowCachedMaxAge in js-sdk

yolken-airplane:
## Description
This change updates the task config schema in the js-sdk to support setting `allowMaxCachedAge`. It also adds a `allowCachedMaxAge ` client option so that this can be customized on a per-task-execution basis. 

See https://www.notion.so/Server-side-task-caching-a1e78a9b426e4456929827a05437a440 for more details.

## Testing
Testing completed successfully via deploying locally.


ameyakhare (approved):

yolken-airplane:
@ameyakhare updated this change to also allow setting `allowCachedMaxAge` in execute calls. Can you take another look? Thanks!

packages/airplane/internal/api/client.ts
```diff
@@ -142,6 +146,9 @@ export class Client {
             slug: slug,
             paramValues: paramValues ?? {},
             resources: resources ?? {},
+            ...(typeof this.allowCachedMaxAge !== "undefined" && {
```
yolken-airplane, line 149: Not using `this.allowCachedMaxAge && ...` since we want to set the value if it's `0`.

ameyakhare: why do you have to do typeof? can't you just check to see if the value is undefined or null?

ameyakhare (commented):
1 comment

packages/airplane/internal/api/client.ts
yolken-airplane: That's what stackoverflow said :-). In any case, I can just set it directly.

yolken-airplane: Done.

PR #466, opened 2023-11-14
Add actions and row actions to js sdk config

rose-huang:
Test plan:
Yarn build and use in ts task

image


packages/airplane/internal/config.ts
```diff
@@ -117,7 +117,100 @@ export type K8SPodSpecPatch = {
...
+type EntityActionBase = ActionBase & {
+  /**
+   * The parameter values when previewing the entity.
+   *
+   * The values can be a JST that can reference the output of the task or
+   * the row of the output if using a row action.
+   */
+  paramValues?: InputParamValues;
+  /**
+   * Method of previewing the entitiy. Defaults to "side_peek".
```
leeweisberger, line 153: Let's use @default rather than having the default in the comment

example: https://github.com/airplanedev/views/blob/6d499c229530476b36d45b94bbb05a125ffa4d5c/lib/src/components/chip/Chip.types.ts#L15

packages/airplane/internal/config.ts
```diff
@@ -117,7 +117,100 @@ export type K8SPodSpecPatch = {
...
+   */
+  runbook: string;
+};
+
+type Action = HrefAction | TaskAction | PageAction | ViewAction | RunbookAction;
+
+type OutputDisplayBase = {
+  /**
+   * Buttons to display alongside the output display. Can be used to open a link or
+   * to open a task, page, view, or runbook in a peek.
```
leeweisberger, line 206: a peek or not depending on the "action_as". We can probably just leave this out.

packages/airplane/internal/config.ts
```diff
@@ -211,6 +304,13 @@ type TableOutputDisplay = {
    * @example "{{ output.map((o) => ({ name: o.name, address: o.address })) }}"
    */
   value?: string;
+  /**
+   * Buttons to display for each row of the table output display. Can be used to open a link or
+   * to open a task, page, view, or runbook in a peek.
```
leeweisberger, line 309: remove "in a peek"

leeweisberger (approved):

PR #470, opened 2023-11-27
Allow clients to get isCached when executing runs

yolken-airplane:
## Description
This change updates the js-sdk runtime interface to support a new method, `executeBackgroundWithCacheInfo`, that's like `executeBackground` but returns an `isCached` value in addition to the `runID` of the created run. The latter is needed for views so we can indicate whether a child run was cached in the activity panel. It may also be helpful for other, user-specific use cases. 

As an alternative to this, we could change the response type to `executeBackground`, but that would break backwards compatibility, which seems a little painful at this point.

This change also updates the sdk to set `inputsZoneParamValuesHash` when creating a run with self-hosted inputs. This extra parameter ensures that cache keys calculated by the API will change if the real input parameter values change in the self-hosted case.

## Testing
Testing completed successfully via deploying and running locally in conjunction with the changes in https://github.com/airplanedev/airport/pull/5948.


ameyakhare (approved):
lgtm

